PR #163, opened 2012-09-05
Prevent SEGFAULT of XmlNodes after they are detached from the tree

skabbes:
[This](https://github.com/polotek/libxmljs/commit/4f7255d4c5d1c8e2a393560603def146cf758186) commit introduced a bug which allows detached XmlNodes to be double free'd.  I've done some research, and this commit is definitely necessary since `xmlUnlinkNode` does **not** free that memory, and neither does `xmlFreeDoc`.  However, we need a mechanism to prevent this behavior, since it is very difficult for end users to debug segfaults (no stack traces provided by Node).

Repro Steps are laid out in the included test case, but basically go like this:

1. Access children of node `a` (so that those nodes are being reference tracked by v8)
2. Remove node `a` from the dom tree (which also remove all children nodes, including attributes)
3. Trigger a garbage collection after all those variables are no longer in scope
4. Calling xmlFreeNode in `a`'s destructor recursively destroys the xml tree of its children (since `a->parent == NULL`)
5. Calling xmlFreeNode on a's children cause them to be freed again (since `child->parent != NULL`)
6. However, `xml_obj->doc->_private` is no longer a valid reference to an XmlDocument, so the unref causes a SEGFAULT

### Possible Solutions

* Simply check `xml_obj->type == -1`, since libxml [automatically resets all memory to -1 after it free's it](http://git.gnome.org/browse/libxml2/tree/xmlmemory.c?id=v2.7.8#n438).  However, this is very much implementation dependent and doesn't allow us to property `unref` the parent document (since the doc pointer is also overwritten)

* Store a pointer to the xmlDoc on the XmlNode object and compare that to `xml_obj->doc` before calling xmlFreeNode on that node.  This assumes that the xmlDoc will never be free'd before any of its children nodes (should be correct).  It also means that we need to maintain that pointer for every operation which may change the node's document (addPrevSibling, addNextSibling off the top of my head)

* Keep track of which nodes have been removed in XmlDocument, and free them explicitly when ~XmlDocument is called.  The code would have to be smart enough to remove nodes from this mapping when they are re-added to the tree.

* Immediately call xmlUnlinkNode and xmlFreeNode inside `XmlNode::remove` and make the api enforce that you cannot reuse any removed node or any of its children.

I'm not super familiar with libxml or v8 or what's possible with them, so maybe someone here who understands the problem can weigh in @shtylman?

### DIFF
--- a/test/ref_integrity.js
+++ b/test/ref_integrity.js
@@ -25,3 +25,26 @@ module.exports.references = function(assert) {
     assert.equal("child", nodes[1].name());
     assert.done();
 };
+
+// test that double-freeing XmlNode's doesn't cause a segfault
+module.exports.double_free = function(assert) {
+    var children = null;
+
+    // stick this portion of code into a self-executing function so
+    // its internal variables can be garbage collected
+    (function(){
+      var html = '<html><body><div><span></span></div></body></html>';
+      var doc = libxml.parseHtml(html);
+
+      doc.find('//div').forEach(function(tag){
+        // provide a reference to childNodes so they are exposed as XmlNodes
+        // and therefore subject to V8's garbage collection
+        children = tag.childNodes();
+        tag.remove();
+      });
+    })();
+
+    global.gc();
+    assert.ok( children[0].attrs() );
+    assert.done();
+};


DullReferenceException:
Is there any sort of workaround we can employ in JS for this bug? I thought I could maybe avoid the double-freeing if I removed child nodes that I'm referencing from the parent node before removing that parent, but I'm still getting the segfault. Essentially:

```javascript
var element = getSomeElement();

var childAttr = element.attr('attrname');
var parent = element.parent();

doSomethingWith(element.text());
doSomethingWith(childAttr.value());

element.text('');
childAttr.remove();
element.remove();
parent.remove();
```

Is there a reason this workaround doesn't work? Is there one that will?

hildjj:
This issue should not be freed until a patch is accepted into master.

rc0x03:
Would it be possible to avoid complicated ref counting code and instead make it the user's responsibility to free the document via the API? When the user knows that they're done adding/creating elements and modifying/searching the document etc., they could simply call `doc.free()`. Is this at all a possibility? Could it be potentially dangerous?

skabbes:
@rc0x03 I don't think that's a good idea from an API perspective.  The memory management code shouldn't be that complex really - someone just needs to take the time and do it carefully.

rc0x03:
That's a good point. Another reason why that might be a bad idea is if the user has complicated async code. The user would then have to determine when they're completely done using the document. After looking through the source I came up with a possible solution.

Libxml has a few function that accept callbacks. One of those functions is [`xmlDeregisterNodeDefault]`(http://www.xmlsoft.org/html/libxml-globals.html#xmlDeregisterNodeDefault). This function set the default callback function that is called when any node is "deregistered" (i.e. before freeing).

We can use this function to prevent bad access related segmentation faults by setting a flag saying that the xmlNode has already been freed. The XmlNode destructor code will then check the flag and return if it has already been freed;

In `libxmljs.cc`:
```

// set up our callback function that will set the flag

void cleanupNode(xmlNode* node) {
  if (node->_private) {
    (static_cast<XmlNode*>(node->_private))->isFree = true;
  }
  return;
}

// set the callback function to be called

LibXMLJS::LibXMLJS()
{
    xmlDeregisterNodeDefault(cleanupNode);

....

}
```

In `xml_node.h`:

```
class XmlNode : public node::ObjectWrap {
public:
    bool isFree;
```

In `xml_node.cc`:
```
// The constructor will set the default value to false

XmlNode::XmlNode(xmlNode* node) : xml_obj(node) {
    this->isFree = false;

...
}


// The destructor will check the value

XmlNode::~XmlNode() {
    if (this->isFree) {
      return;
    }

...
}
```

rc0x03:
Note: this only fixes cases where a parent destructor gets called before a child destructor and the child node is double freed.

This does not fix the problem in pull request #267. The problem here is that the parent is allowed to be freed even though there is still a floating reference to one of its child nodes.

rc0x03:
This seems to be the same way ruby does it:
https://github.com/xml4r/libxml-ruby/blob/master/ext/libxml/ruby_xml_node.c

rc0x03:
Pull request #323 solves this issue.

PR #267, opened 2014-10-21
Fix double free

skabbes:
Partially fixes: https://github.com/polotek/libxmljs/pull/163

After some exploration, it appears that `xmlFreeNode` actually recursively frees all children nodes.  If those children nodes have ObjectWrap'd wrappers pointing to them, we then have dangling `xml_obj`'s.  This PR addresses this by freeing those nodes only up to any nodes with `xml_obj->_private != NULL`.  This effectively promotes those nodes to the be the new "owners" of their subtrees.

Although this fixes the ref_integrity tests isolation, it seems to expose some more bugs in test/attribute.js.  I haven't been able to track those down yet.

### DIFF
--- a/src/xml_node.cc
+++ b/src/xml_node.cc
@@ -196,8 +196,9 @@ XmlNode::~XmlNode() {
 
     // We do not free the xmlNode here if it is linked to a document
     // It will be freed when the doc is freed
-    if (xml_obj->parent == NULL)
-      xmlFreeNode(xml_obj);
+    if (xml_obj->parent == NULL) {
+      XmlNode::free_untracked_child_nodes(xml_obj);
+    }
 }
 
 v8::Local<v8::Value>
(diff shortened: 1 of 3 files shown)


defunctzombie:
@skabbes is this something we should merge in even if a partial solution?

skabbes:
The fix is pretty good, I think it would be good to get into master.  However, apparently some tests started failing in `test/attribute.js`.  Perhaps we should identify why that is failing before merging.

defunctzombie:
Yep, we can't merge if tests are not passing.

rc0x03:
A possible solution to this would be to give all newly created ObjectWrapped XmlNodes a reference to their "creator" ObjectWrapped XmlNode.

The constructor for any new the XmlNode would then `ref` the "parent" ObjectWrap, keeping it from being GCed before the child.

The destructor for new XmlNodes would then `unref` the "parent" ObjectWrap, allowing the child node to be freed.

Once the parent ObjectWrapped XmlNode's ref count goes down to 0, V8 will then free it. This will ensure that nodes are only freed once all of their children have been freed.

rc0x03:
This seems to be the same way ruby does it:
https://github.com/xml4r/libxml-ruby/blob/master/ext/libxml/ruby_xml_node.c

rc0x03:
Pull request #323 seems to be an effective solution to this issue.

teleological:
#323 was merged, so this PR could be closed out.

PR #271, opened 2014-11-03
Add .nsDecls API

### LINKED ISSUE #270: No API for namespace declarations
Consider the following code:

```javascript
var libxmljs = require("libxmljs");

var xmlstr = '<foo xmlns:b="urn:example:1"><bar xmlns:c="urn:example:2"/></foo>'
var xmlDoc = libxmljs.parseXmlString(xmlstr);

var bar = xmlDoc.get('bar');
bar.namespaces().forEach(function(n) {
  console.log(n.prefix(), n.href());
});
```

The output:
```
c urn:example:2
b urn:example:1
```

The `namespaces()` API is useful to see what namespaces are in scope on a particular element, but it is useless to tell what namespaces are *declared* on an element.  This is important if you wanted to walk a node recursively, recreating the input exactly, without having to parse the difference between `namespaces()` calls on the current element and its parent.

I suggest adding a `namespaceDeclarations()` or `nsDecls`API that makes this available.  I'm happy to work up a patch if this makes sense to you.

hildjj:
Fixes #270.  New API:

```javascript
  var str = '<html xmlns="urn:example" xmlns:ex1="urn:example:1"><body/></html>';
  var doc = libxml.parseXmlString(str);
  var root = doc.root();
  var decls = root.nsDecls()
```

`decls` now includes two namespace objects.  Applying to the body element yields an empty array.

Tests included.  I'll add docs to the wiki if you take the patch.  I'm open to renaming the API anything you like.

### DIFF
--- a/src/xml_node.cc
+++ b/src/xml_node.cc
@@ -91,6 +91,14 @@ NAN_METHOD(XmlNode::Namespaces) {
   NanReturnValue(node->get_all_namespaces());
 }
 
+NAN_METHOD(XmlNode::NamespaceDeclarations) {
+  NanScope();
+  XmlNode* node = ObjectWrap::Unwrap<XmlNode>(args.Holder());
+  assert(node);
+
+  NanReturnValue(node->get_nsdefs());
+}
+
 NAN_METHOD(XmlNode::Parent) {
   NanScope();
   XmlNode* node = ObjectWrap::Unwrap<XmlNode>(args.Holder());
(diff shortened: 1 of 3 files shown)


hildjj:
This PR does still work.  Would you mind either merging this, giving me something to change, or proposing another approach for #270?

teleological:
Is there a use case to prefer this to `namespaces()` i.e. `xmlGetNsList` ? In any case the node will need to be type checked since attributes don't include an `nsDef` member.

hildjj:
`.namespaces()` gives the set of namespaces that are currently in scope for a given node.  For the example given above:

```
var body = root.childNodes(0);
body.namespaces();
```

still yields two namespaces.  If you're iterating through a document, keeping track of the current state of the namespace stack just to watch for changes is somewhat painful.

Returning to this code, I kind of hate the string `nsDecls`, and wouldn't mind changing it.  I'll be happy to add a node check if there's a shot you're going to take this.

rc0x03:
Maybe we can make it `namespace([local])` and `namespaces([local])` instead, where `local` is an optional boolean argument specifying if we should return all namespaces or local namespace declarations.

I think we could merge this if API is more integrated and the NAN calls are updated to version 2.

hildjj:
Started a new PR at #363, with the suggested API.

PR #273, opened 2014-11-07
Added support for RELAX NG schema validation

rprinz08:


defunctzombie:
@polotek you good with this? I am.

polotek:
lgtm

PR #279, opened 2015-01-16
update nan to 1.5.1 and fix issues with the latest v8

jhs67:
This updates nan to the latest and fixes some issues so that the module can work with the io.js 1.0.1 release.

Most changes are pretty trivial. sax_parser.js had to change the way it inherits from EventEmitter because the old way would cause an assertion trying to set up the ObjectWrap.

### DIFF
--- a/src/xml_comment.cc
+++ b/src/xml_comment.cc
@@ -82,7 +82,7 @@ XmlComment::get_content() {
     return NanEscapeScope(ret_content);
   }
 
-  return NanEscapeScope(NanNew<v8::String>());
+  return NanEscapeScope(NanNew<v8::String>(""));
 }
 
 
@@ -109,7 +109,7 @@ void
 XmlComment::Initialize(v8::Handle<v8::Object> target)
 {
     NanScope();
-    v8::Local<v8::FunctionTemplate> t = NanNew<v8::FunctionTemplate>(New);
+    v8::Local<v8::FunctionTemplate> t = NanNew<v8::FunctionTemplate>(static_cast<NAN_METHOD((*))>(New));
     t->Inherit(NanNew(XmlNode::constructor_template));
     t->InstanceTemplate()->SetInternalFieldCount(1);
     NanAssignPersistent(constructor_template, t);
--- a/lib/sax_parser.js
+++ b/lib/sax_parser.js
@@ -1,5 +1,4 @@
 var events = require('events');
-var util = require('util');
 
 var bindings = require('./bindings');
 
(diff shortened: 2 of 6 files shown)


defunctzombie:
util.inherits does not work? Will IOjs be fixing that?

jhs67:
It isn't that util.inherits doesn't work in io.js, it is that it doesn't (seem) to work for `ObjectWrap` types. I suspect it has more to do with v8 than with anything io.js is doing. [Here's](https://groups.google.com/forum/#!topic/nodejs/U816Fqajn0s) a discussion about it, with a link to the solution I [borrowed](https://github.com/bnoordhuis/node-event-emitter/blob/master/event-emitter.js).

With the old solution, when you replace `bindings.SaxParser.prototype` with a new object (this is what `util.inherits` does) it asserts. In particular, when you call the constructor it routes through to `XmlSaxParser::NewParser`, but when this function calls `ObjectWrap::Wrap` then it asserts because the `args.Holder()` object doesn't have enough `InternalFieldCount` (even though you call `SetInternalFieldCount` in the `XmlSaxParser::Initialize` function). I guess the internal field count information is being saved in the prototype and gets wiped out when the prototype is replaced.

src/xml_comment.cc
```diff
@@ -82,7 +82,7 @@ XmlComment::get_content() {
     return NanEscapeScope(ret_content);
   }
 
-  return NanEscapeScope(NanNew<v8::String>());
+  return NanEscapeScope(NanNew<v8::String>(""));
```
kkoopa: The empty string constructor was accidentally omitted in the NanNew rewrite. It will be restored in the next patch release (next week). https://github.com/rvagg/nan/pull/237

src/xml_comment.cc
```diff
@@ -109,7 +109,7 @@ void
 XmlComment::Initialize(v8::Handle<v8::Object> target)
 {
     NanScope();
-    v8::Local<v8::FunctionTemplate> t = NanNew<v8::FunctionTemplate>(New);
+    v8::Local<v8::FunctionTemplate> t = NanNew<v8::FunctionTemplate>(static_cast<NAN_METHOD((*))>(New));
```
kkoopa: This is somewhat odd. A cast should not be necessary.

jhs67: There are two overloaded New methods and it gave an error about being unable to pick which one. There is probably a new overload of NanNew<FunctionTemplate> that is causing the problem. Alternatively one of the methods could be renamed.

kkoopa: Yes, you are right about that. It cannot tell which one is meant to be used, as both are in the same namespace, although it would not make sense to use the method that does not fit the signature. I'll see if this can be improved somehow.

On a side note, that other New method does not seem to be used at all. Removing it produces no errors.

kkoopa: Should be fixed in next patch release.
rvagg/nan@5ef813bf06c93725d03d17a91ee48d2c66dd7b3b

jhs67: Nice, thank you. I'll submit a patch cleaning things up when the new release is available.

Issue #201, opened 2013-06-18
Proper Parsing/Writing attributes with namespaces


kcgolden:
Hi,

I am trying to read/write attributes from AndroidManifest.xml but the attributes are usually prefixed with 'android:'

Example:
~~~~~~~~
<supports-screens android:xlargeScreens="true" android:largeScreens="true" android:normalScreens="true" android:smallScreens="true" android:resizeable="true" android:anyDensity="true"/>
~~~~~~~~
I can read the attribute with:

xmlDoc.attr('anyDensity').value();

I couldn't find it if I did:

xmlDoc.attr('android:anyDensity').value();

The second option is more explicit so that would be preferable. I also need to be able write new attributes with the 'android:' prefex and I haven't a clue how to do that.

I "think" this is a feature request, if not a bug since we don't seem to be able to have attributes of the form "string:string". However, I am willing to accept the possibility that I'm missing something fundamental and it is possible to do what I desire with libxmljs using some vodoo I have not yet discovered. Can you help me out?

benoror:
Updated the wiki accordingly: https://github.com/polotek/libxmljs/wiki/Attribute#attributenamespace

defunctzombie:
Closing this as I think the namespace method on attribute gets the result needed. If that is not the case we can re-open.

PR #284, opened 2015-01-28
Add .namespace() to Attribute

katoken-0215:
Proposal to #201 

This is useful to treat attribute with namespace.

Example:
```javascript
var libxmljs = require('./index.js');

var xml = '<root ' +
      'xmlns:foo="http://www.example.com/fake/uri/foo"' +
      'xmlns:bar="http://www.example.com/fake/uri/bar"' +
      '>' +
      '<node foo:attr="foo value" bar:attr="bar value" />' +
      '</root>';

var doc = new libxmljs.parseXmlString(xml);
var node = doc.get('node');
var attrs = node.attrs();

attrs.forEach(function(attr) {
  console.log(attr.namespace().prefix() + ':' + attr.name() + '="' + attr.value() + '"');
});
```

Output:
```
foo:attr="foo value"
bar:attr="bar value"
```

### DIFF
--- a/src/xml_attribute.cc
+++ b/src/xml_attribute.cc
@@ -73,6 +73,14 @@ NAN_METHOD(XmlAttribute::Node) {
   NanReturnValue(attr->get_element());
 }
 
+NAN_METHOD(XmlAttribute::Namespace) {
+  NanScope();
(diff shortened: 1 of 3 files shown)


defunctzombie:
LGTM

PR #297, opened 2015-04-04
Vendor submodule

gagern:
Instead of copying the libxml source code into this repository, it might be better to use a git submodule instead. In particular doing so will make it easier to update to a more recent version of libxml (say 2.9.2) since all one has to do is cd into the `vendor/libxml` directory, do a `git checkout 2.9.2` there, and add that change to a commit in the base repo. It also makes the whole history of the libxml2 project available for developers, which can be useful at times. And it keeps the libxmljs repo smaller, since we won't have to store much information while following upstream development.

This change does the switch from copied code to submodule, but it leaves the version as is, at 2.9.0. Any files we had in use which are not included in the upstream tree were moved to a different location. So unless I messed something up, this whole change should have zero impact on the end users, although it might be a bit confusing for developers who don't check out submodules by default. A simple `git submodule init` followed by `git submodule update` will take care of that.


defunctzombie:
I don't think it is necessary. You can just copy over the existing files to update libxmljs which we don't do all that often anyway.

I get that this is trying to use submodules in place of any common module system for c/c++ code but really the libxmljs source is part of the binding and I think of it all as one unit.

Klortho:
👎 This should have been pulled.

PR #299, opened 2015-04-06
Make NanAdjustExternalMemory thread-safe

### LINKED ISSUE #296: Async use of NanAdjustExternalMemory causes crash with Node 0.12.1
I'm using node 0.12.1 and am investigating a segmentation fault in node-libxslt which builds on libxmljs. See albanm/node-libxslt#9.

I found that the segmentation fault occurs in a worker thread (i.e. not the V8 event dispatcher thread) when that thread runs some libxml code, and as a consequence calls `NanAdjustExternalMemory` at some point. In that thread, `v8::Isolate::GetCurrent()` will return `NULL`, causing the crash. The way I understand it, calls to V8 API from threads other than the event dispatcher thread are forbidden, including that Nan call.

Unfortunately I still have to find a good documentation on that `AdjustAmountOfExternalAllocatedMemory` method of the isolate, to see what repercussions we'd have to expect if we simply stopped telling V8 about our memory usage. But I guess 393b44a0010da5a469222291e9cad6be74f33a4c introduced that code for a reason. That commit is titled “first round of fixes for external memory issues” but doesn't say what kinds of issues these are. @polotek, do you remember any details?

If there is a really strong reason to keep these functions, then we'll probably have to work out how to make them function in an asynchroneous world. Implement some libuv signalling which reports adjustments from worker threads back to the event dispatch thread to be fed to V8. I very much hope that we don't really need this, since it might add considerable synchronization overhead.

nathan7:
Fixes #296.
WIP: currently causes the tests to hang on exit.

### DIFF
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -1,6 +1,7 @@
 // Copyright 2009, Squish Tech, LLC.
 
 #include <v8.h>
+#include <uv.h>
 
 #include <libxml/xmlmemory.h>
 
@@ -17,31 +18,11 @@ LibXMLJS LibXMLJS::instance;
 
 // track how much memory libxml2 is using
 int xml_memory_used = 0;
+int xml_memory_used_last = 0;
+uv_async_t xml_memory_handle;
+uv_mutex_t xml_memory_mutex;
 
-// wrapper for xmlMemMalloc to update v8's knowledge of memory used
-// the GC relies on this information
-void* xmlMemMallocWrap(size_t size)
-{
-    void* res = xmlMemMalloc(size);
-
-    // no need to udpate memory if we didn't allocate
-    if (!res)
-    {
-        return res;
-    }
-
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
-    return res;
-}
-
-// wrapper for xmlMemFree to update v8's knowledge of memory used
-// the GC relies on this information
-void xmlMemFreeWrap(void* p)
-{
-    xmlMemFree(p);
-
+NAUV_WORK_CB(xml_memory_cb) {
     // if v8 is no longer running, don't try to adjust memory
     // this happens when the v8 vm is shutdown and the program is exiting
     // our cleanup routines for libxml will be called (freeing memory)
@@ -58,9 +39,49 @@ void xmlMemFreeWrap(void* p)
         return;
     }
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
+    uv_mutex_lock(&xml_memory_mutex);
+    const int diff = xml_memory_used - xml_memory_used_last;
+    xml_memory_used_last = xml_memory_used;
+    uv_mutex_unlock(&xml_memory_mutex);
+
     NanAdjustExternalMemory(diff);
+}
+
+NAN_INLINE void xml_memory_update() {
+    uv_mutex_lock(&xml_memory_mutex);
+
+    xml_memory_used = xmlMemUsed();
+    if (xml_memory_used != xml_memory_used_last) {
+        uv_async_send(&xml_memory_handle);
+    }
+
+    uv_mutex_unlock(&xml_memory_mutex);
+}
+
+// wrapper for xmlMemMalloc to update v8's knowledge of memory used
+// the GC relies on this information
+void* xmlMemMallocWrap(size_t size)
+{
+    void* res = xmlMemMalloc(size);
+
+    // no need to udpate memory if we didn't allocate
+    if (!res)
+    {
+        return res;
+    }
+
+    xml_memory_update();
+
+    return res;
+}
+
+// wrapper for xmlMemFree to update v8's knowledge of memory used
+// the GC relies on this information
+void xmlMemFreeWrap(void* p)
+{
+    xmlMemFree(p);
+
+    xml_memory_update();
 }
 
 // wrapper for xmlMemRealloc to update v8's knowledge of memory used
@@ -74,9 +95,8 @@ void* xmlMemReallocWrap(void* ptr, size_t size)
         return res;
     }
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
+    xml_memory_update();
+
     return res;
 }
 
@@ -91,14 +111,17 @@ char* xmlMemoryStrdupWrap(const char* str)
         return res;
     }
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
+    xml_memory_update();
+
     return res;
 }
 
 LibXMLJS::LibXMLJS()
 {
+
+    uv_mutex_init(&xml_memory_mutex);
+    uv_async_init(uv_default_loop(), &xml_memory_handle, xml_memory_cb);
+    uv_unref((uv_handle_t*) &xml_memory_handle);
 
     // populated debugMemSize (see xmlmemory.h/c) and makes the call to
     // xmlMemUsed work, this must happen first!
@@ -109,7 +132,7 @@ LibXMLJS::LibXMLJS()
     LIBXML_TEST_VERSION;
 
     // initial memory usage
-    xml_memory_used = xmlMemUsed();
+    xml_memory_used = xml_memory_used_last = xmlMemUsed();
 }
 
 LibXMLJS::~LibXMLJS()


nathan7:
This could probably be cheaper than a mutex, using atomics, but this is the simplest solution that works. I'm not entirely sure yet why it hangs.

nathan7:
This also seems to fail to compile on node, thanks to the libuv `uv_async_cb` signature changing across versions. Gotta sort that one out.

nathan7:
It appears `NAUV_WORK_CB` could do the trick for that, working on that now.

polotek:
Thanks for the start. After reading this about how libuv wants to treat threads, I'm not sure things will be so cut and dry. https://github.com/nikhilm/uvbook/blob/master/source/threads.rst

The real problem is that what we need to do here is update the *current* memory usage with v8. But because these callbacks are asynchronous, and libuv might even *combine* multiple callbacks, we are left in a position where the current memory usage might have changed by the time the callback fires.

So we could explore a few different things:

1. Keep track of multiple allocations across threads and then reconcile them whenever a callback eventually fires. This doesn't sound great to me.
2. Run the memory checker on some reasonable interval and then update v8 with whatever the current libxml2 memory usage is. This will be less accurate than hints that happen at the point of allocation. But it sounds easier to implement and might still provide most of the benefit.

Does that seem to make sense? I think what you're doing is similar to option 2. But I'm not sure it's complete. For one thing, I'm not sure how `xml_memory_used_last` is being used. It seems to only be updated on initialization. I could be missing something though. Could you give some info on your approach?

nathan7:
Oh, welp! I was totally forgetting to update the `_last` var after sorting everything.
`xml_memory_used_last` keeps track of the last usage level V8 is aware of, and `xml_memory_used` keeps track of the current usage level.
Whenever the callback fires, it matches the two up. The V8 usage ends up being eventually consistent, but that's the best we can do here.

nathan7:
Seems the hang isn't because V8 still believes there's memory to be reclaimed ­— I'm not sure what's going on here.
```
(gdb) print libxmljs::xml_memory_used - libxmljs::xml_memory_used_last 
$1 = 0
```

nathan7:
Found it!
```
(gdb) print uv_default_loop().handle_queue 
$3 = {0x1a4df08 <default_loop_struct+648>, 0x7f5d7d6be420 <libxmljs::xml_memory_handle+32>}
```

nathan7:
Tests are broken on 0.8 because npm is broken on 0.8 due to iarna/are-we-there-yet#3. Other than that, we're golden!

nathan7:
I've tested this on 0.8 with a working npm, and all tests pass successfully.

polotek:
The Travis build is failing. I'll look into this a bit later. Thanks for the work @nathan7.

Just to be clear, I think we're on the same page about the memory tracking being "eventually consistent" with this approach. It's similar to #2 that I described above. But instead of running on an interval, it runs every time a callback is scheduled. This is better then an interval in the sense that you can avoid doing this update if memory needs haven't changed. But it has the downside that multiple callbacks may get scheduled and/or combined. And so it becomes much less deterministic when this runs and with what frequency. I'm not sure what the implications of that are.

I'd like to compare this change with what we had before under heavy memory usage patterns. Meanwhile if @albanm wants to try out this patch and let us know if it works under node-libxslt, that would be helpful. Thanks all.

nathan7:
@polotek The reason Travis tests are failing is documented above. This is due to an npm bug, and I'm working on getting that fixed.
The callbacks being combined is an explicit goal here, it reduces the cost of keeping V8 up to date significantly. As far as I can tell, the combining is guaranteed from the libuv side, though the code handles it correctly if that isn't the case.
The present behaviour is absolutely incorrect — it invokes undefined behaviour, which currently seems to visibly manifest as segfaults, but it may also cause arbitrary memory corruption or worse.
(undefined behaviour includes anything from a fairly benign segfault to doing unspeakable acts of perversion with your pet dachshund, after all)
We could try a hybrid approach, but I think event loop latency functions as an effective enough timer.
The only possibly worthwhile optimisation I can see is using atomics to keep the tally, to remove the cost of the mutex in allocation/deallocation. I think that's actually fairly minor, but I'm open to profiler evidence that I'm wrong.

polotek:
What I meant by comparing the current behavior with this patch is making sure we understand what the different resource usage looks like. This current behavior was specifically to address people's concerns with libxmljs's memory profile under heavy usage.

I want to make sure I understand what you mean by "the current behavior is absolutely incorrect". My understanding is this isn't a case of incorrect code. This is a case of non-thread safe code being used in a threaded context. This problem doesn't happen in libxmljs. It happens in other modules that use libxmljs in a way it wasn't designed for.

Do you understand where I'm coming from. I'm not opposed to this change. But I want to make sure we do due diligence before landing it.

polotek:
Also, I misunderstood your comment about build failures due to npm versions. I thought it was just an issue you were dealing with. I didn't realize that was also breaking our travis builds. Carry on.

nathan7:
@polotek 

> What I meant by comparing the current behavior with this patch is making sure we understand what the different resource usage looks like. This current behavior was specifically to address people's concerns with libxmljs's memory profile under heavy usage.

I do understand where you're coming from, and I appreciate your caution.


> I want to make sure I understand what you mean by "the current behavior is absolutely incorrect". My understanding is this isn't a case of incorrect code. This is a case of non-thread safe code being used in a threaded context. This problem doesn't happen in libxmljs. It happens in other modules that use libxmljs in a way it wasn't designed for.

This is incorrect, as far as I can tell, even in the most simple case.
At least two threads are involved, the V8 thread, and the libxmljs worker thread.
The `NanAdjustExternalMemory` call runs on the worker thread, and is unsynchronised with the V8 thread. The structures it accesses are protected with the isolate's V8 mutex — whatever thread is holding it is the V8 thread. V8 expects to be able to freely manipulate those structures on both sides (regular V8 usage, and our `NanAdjustExternalMemory`), because it believes it is holding that mutex. Accessing or modifying these structures without holding the mutex invokes undefined behaviour, which manifests as arbitrary memory corruption, or in the most benign case, a segfault.

This is a problem even in the most basic usage of libxmljs. It will manifest less in I/O-bound programs (because V8 isn't running as much of the time, and thus isn't modifying or accessing those structures), and more in CPU-bound programs, specifically those heavy on allocation.
This bug is hence probabilistic in nature, but can probably be reproduced relatively easily.

I don't believe arbitrary memory corruption is to be taken lightly, and we're lucky someone narrowed their segfaults down to being caused by libxmljs. It could have introduced arbitrary behaviour to other components of their system.

> What I meant by comparing the current behavior with this patch is making sure we understand what the different resource usage looks like. This current behavior was specifically to address people's concerns with libxmljs's memory profile under heavy usage.

Because we were manipulating the piece of memory that holds V8's external memory size without synchronisation, some of our modifications may be overwritten, or we may overwrite V8's modifications (or those from other node extensions) to falsely cause it not to account for memory it has freed or allocated.

```c
// V8 thread
v8_magic_structure->memoryMagic->externalMemory += allocationSize;
// our thread
v8_magic_structure->memoryMagic->externalMemory -= ourAllocationSize;

// a troublesome interleaving, showing the actual load/stores involved:
int temp = v8_magic_structure->memoryMagic->externalMemory;
int ourTemp = v8_magic_structure->memoryMagic->externalMemory;
v8_magic_structure->memoryMagic->externalMemory = temp + allocationSize;
v8_magic_structure->memoryMagic->externalMemory = ourTemp - ourAllocationSize;
```

The new count may lag behind by an event loop tick, but it will never corrupt the global count for the process's entire lifetime.

On other platforms than x86, this bug might manifest more frequently — we're just lucky enough that pointer-sized stores on x86 are atomic. Lacking any memory barriers (such as the ones the mutex introduces), the compiler will of course assume the variable doesn't change.

polotek:
At least two threads are involved, the V8 thread, and the libxmljs worker thread.

I think this is where my confusion lies. I may be showing my ignorance of libxml2 and native node bindings here. But I'm okay with that because I never claimed to be an expert :)

My current understanding is that in the normal case of running node with libxmljs, there is no "libxmljs worker thread". v8 functions and libxml2 functions run in the same single process thread. And when you call into libxmljs, you are blocking the main event loop. What I think I hear you saying is that even using libxmljs in the normal way, and not explicitly pushing it into worker threads, that libxml2 creates it's own worker thread under the hood. That is big news to me.

Libxmljs doesn't currently support asynchronous actions. Even the file reading was synchronous before we removed it entirely and just told people they could load their own file data using standard node fs methods. If what you're saying is true (and not just a consequence of how node-libxslt is trying to use it), then I need to make sure I get a better handle on this issue. I'm going to dig into it myself. But if you could point me at places where I can read up on this, that would be helpful.

nathan7:
@polotek Okay, if you're entirely right and there are no other threads involved, that means I'm fixing a bug that doesn't exist, and the original issue was wrong entirely. But I'm not sure that is the case.

nathan7:
I did some science, making the two parts being called from different threads cause an `abort()` call:
```
 ~  src  github.com  polotek  libxmljs  thread-safe  npm test

> libxmljs@0.13.0 test /home/nathan/src/github.com/polotek/libxmljs
> node --expose_gc ./node_modules/.bin/nodeunit test

xmlMemSetup()
xmlMemSetup() Ok
xmlInitMemory()
xmlInitMemory() Ok
Malloc(400)
Malloc(400) Ok
fish: Job 1, “npm test” terminated by signal SIGABRT (Abort)
```

Patch included, so anyone can reproduce this:
```diff
diff --git a/src/libxmljs.cc b/src/libxmljs.cc
index dd03395..7f21d57 100644
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -22,6 +22,15 @@ int xml_memory_used_last = 0;
 uv_async_t xml_memory_handle;
 uv_mutex_t xml_memory_mutex;
 
+pthread_t thread;
+void check_thread() {
+    pthread_t self = pthread_self();
+    if (!pthread_equal(self, thread)) {
+        abort();
+    }
+    thread = self;
+}
+
 NAUV_WORK_CB(xml_memory_cb) {
     // if v8 is no longer running, don't try to adjust memory
     // this happens when the v8 vm is shutdown and the program is exiting
@@ -42,6 +51,9 @@ NAUV_WORK_CB(xml_memory_cb) {
     uv_mutex_lock(&xml_memory_mutex);
     const int diff = xml_memory_used - xml_memory_used_last;
     xml_memory_used_last = xml_memory_used;
+
+    check_thread();
+
     uv_mutex_unlock(&xml_memory_mutex);
 
     NanAdjustExternalMemory(diff);
@@ -55,6 +67,8 @@ NAN_INLINE void xml_memory_update() {
         uv_async_send(&xml_memory_handle);
     }
 
+    check_thread();
+
     uv_mutex_unlock(&xml_memory_mutex);
 }
 
@@ -144,6 +158,8 @@ LibXMLJS::~LibXMLJS()
 extern "C" void
 init(v8::Handle<v8::Object> target)
 {
+      thread = pthread_self();
+
       NanScope();
 
       XmlDocument::Initialize(target);
```

nathan7:
I did some further testing to try and falsify my hypothesis: @polotek, you're right, and the original issue is wrong. This PR is useless entirely. The original issue mis-identifies the cause.

The following patch does not cause an `abort()`, and all tests run successfully.
My earlier findings were a mistake in my experiment, due to a misunderstanding about what gets executed when.
```diff
diff --git a/src/libxmljs.cc b/src/libxmljs.cc
index dd03395..997a333 100644
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -22,6 +22,19 @@ int xml_memory_used_last = 0;
 uv_async_t xml_memory_handle;
 uv_mutex_t xml_memory_mutex;
 
+pthread_t thread = 0;
+void check_thread() {
+    pthread_t self = pthread_self();
+    if (thread == 0) {
+        thread = self;
+    }
+
+    if (!pthread_equal(self, thread)) {
+        abort();
+    }
+    thread = self;
+}
+
 NAUV_WORK_CB(xml_memory_cb) {
     // if v8 is no longer running, don't try to adjust memory
     // this happens when the v8 vm is shutdown and the program is exiting
@@ -42,6 +55,7 @@ NAUV_WORK_CB(xml_memory_cb) {
     uv_mutex_lock(&xml_memory_mutex);
     const int diff = xml_memory_used - xml_memory_used_last;
     xml_memory_used_last = xml_memory_used;
+    check_thread();
     uv_mutex_unlock(&xml_memory_mutex);
 
     NanAdjustExternalMemory(diff);
@@ -55,6 +69,8 @@ NAN_INLINE void xml_memory_update() {
         uv_async_send(&xml_memory_handle);
     }
 
+    check_thread();
+
     uv_mutex_unlock(&xml_memory_mutex);
 }
```

nathan7:
This *can* be reproduced using the above patch once we bring node-libxslt into the mix — a release build, because the tests are *really* noisy otherwise:
```
 ~/src/github.com/albanm/node-libxslt   master  ⬢  0.12.2  node_modules/.bin/mocha -R spec 
child_process: customFds option is deprecated, use stdio instead.


  node-libxslt
    synchronous parse function
      ✓ should parse a stylesheet from a libxslt.libxmljs xml document 
      ✓ should parse a stylesheet from a xml string 
      ✓ should parse a stylesheet with a include from a xml string 
      ✓ should throw an error when parsing invalid stylesheet 
    parseFile function
zsh: abort (core dumped)  node_modules/.bin/mocha -R spec
```

nathan7:
Results with libxmljs master:
```
 ~/src/github.com/albanm/node-libxslt   master  ⬢  0.12.2  npm test

> libxslt@0.4.0 test /home/nathan/src/github.com/albanm/node-libxslt
> mocha -R spec

child_process: customFds option is deprecated, use stdio instead.


  node-libxslt
    synchronous parse function
      ✓ should parse a stylesheet from a libxslt.libxmljs xml document 
      ✓ should parse a stylesheet from a xml string 
      ✓ should parse a stylesheet with a include from a xml string 
      ✓ should throw an error when parsing invalid stylesheet 
    parseFile function
zsh: segmentation fault (core dumped)  npm test
```

Results with this PR:
```
 ~/src/github.com/albanm/node-libxslt   master  ⬢  0.12.2  npm test

> libxslt@0.4.0 test /home/nathan/src/github.com/albanm/node-libxslt
> mocha -R spec

child_process: customFds option is deprecated, use stdio instead.


  node-libxslt
    synchronous parse function
      ✓ should parse a stylesheet from a libxslt.libxmljs xml document 
      ✓ should parse a stylesheet from a xml string 
      ✓ should parse a stylesheet with a include from a xml string 
      ✓ should throw an error when parsing invalid stylesheet 
    parseFile function
      ✓ should parse a stylesheet from a file 
    asynchronous parse function
      ✓ should parse a stylesheet from a libxslt.libxmljs xml document 
      ✓ should parse a stylesheet from a xml string 
      ✓ should parse a stylesheet with a include from a xml string 
      ✓ should return an error when parsing invalid stylesheet 
    synchronous apply function
      ✓ should apply a stylesheet to a libxslt.libxmljs xml document 
      ✓ should apply a stylesheet to a xml string 
      ✓ should apply a stylesheet with a parameter 
      ✓ should apply a stylesheet with a include to a xml string 
    asynchronous apply function
      ✓ should apply a stylesheet to a libxslt.libxmljs xml document 
      ✓ should apply a stylesheet to a xml string 
    applyToFile function
      ✓ should apply a stylesheet to a xml file 
    disable-output-escaping attribute
      ✓ should be interpreted by a stylesheet 
    libexslt bindings
      ✓ should expose EXSLT functions 


  18 passing (23ms)
```

polotek:
Thanks for this work @nathan7. It's really illustrative and helpful.

defunctzombie:
Is this PR replaced by a newer one? Please close if no longer relevant.

nathan7:
Closing this in favour of #302.

PR #300, opened 2015-04-07
Implement thread-local memory adjustments

gagern:
If the current thread is not the main V8 event thread, we don't report memory adjustments back to V8 immediately, but instead keep track of all the changes in a thread-local variable.  The net change will be reported back after the worker is done.

All state changes are encapsulated in two RAII sentinels: WorkerParent and WorkerSentinel.  The former is constructed and destroyed on the V8 thread, the latter on the worker thread.  Having these in place helps to make client code thread-safe with respect to the state in question.  It also provides a convenient extension point: if, at some point, we need more code for handling worker threads, all we have to do is extend these two classes.

The current delegation to the libxml2 memory management functions has ben completely dropped.  The main use case for these would have been memory debugging, so they do a lot of work (like maintaining linked lists between allocated blocks) that we don't need.  The main thing we could have gained from them, namely the call to xmlMemUsed, was not thread-safe in any case.

With [these adjustments](https://github.com/albanm/node-libxslt/compare/master...gagern:threadLocalMemAdjust), node-libxslt works with this code here.


gagern:
Travis complains, since apparently TLS support in libuv is a more recent thing. To be more precise, joyent/libuv#904 added this for libuv v0.11.11. So for older versions, we'll have to duplicate some of that code, I fear. Except if we can somehow use C++11 and its [`thread_local`](http://en.cppreference.com/w/cpp/language/storage_duration) storage class specifier?

gagern:
OK, Travis is still complaining, but now only for ancient Node 0.8, and I don't see my code being responsible for this. Are you willing to merge my approach?

defunctzombie:
This should work in node 0.8 and you should investigate why that is failing because master works in 0.8 on travis last I checked.

gagern:
Emphasis on “last I checked”. #301 fails despite the change being just one whitespace. See also [this Stack Overflow question](http://stackoverflow.com/q/29494154/1468366) for details.

src/libxmljs.cc
```diff
@@ -50,66 +88,66 @@ void xmlMemFreeWrap(void* p)
...
+}
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
+char* memStrdup(const char* str)
+{
+    size_t size = strlen(str) + 1;
+    char* res = static_cast<char*>(memMalloc(size));
+    if (res) strcpy(res, str);
```
nathan7: This should be a memcpy, given that the string length has already been determined.

gagern:
The Nan people have accepted rvagg/nan#307 so we can drop `uvcompat.h` once Nan 1.8.0 gets released. Until then we could of course depend on a specific github commit, but that feels somewhat unclean. While nathan7 continues in #302 to work towards an async-based solution, I still feel that this approach here will be more flexible solution to #296 in the long run.

defunctzombie:
@kkoopa can you release a new nan?

defunctzombie:
Is this PR different than what @302 fixes?

gagern:
Both fix #296, so both have the same aim. This one here requires the cooperation of whoever sets up worker threads, but on the other hand avoids locking and signaling overhead. I guess it's time to write some demo app which can be used to benchmark the two approaches. Perhaps by simply extending libxmljs with some async parser function, and running that in a loop over some big XML files? I'll have a look at that soon.

src/libxmljs.cc
```diff
@@ -9,39 +9,78 @@
...
+}
+
+struct memHdr {
+    size_t size;
+};
+
+// When we move to C++11 one day, the following would become constexpr,
+// and SLIGN_SIZE could perhaps become sizeof(std::max_align_t).
+#define ALIGN_SIZE sizeof(double)
+#define HDR_SIZE (((sizeof(memHdr) + (ALIGN_SIZE - 1))/ALIGN_SIZE)*ALIGN_SIZE)
```
nathan7: I'm still curious why you choose this over an `offsetof(foo, data)`. I have no idea what's going on with the `HDR_SIZE` macro.
I've refined my approach a little more: https://github.com/nathan7/libshred/blob/dd8591b1e34154166c8cdac513c019d7629f22c7/alloc.h

gagern: This is based on [what libxml does](https://github.com/GNOME/libxml2/blob/38bbd3412d1c2a2016cf5a1a221ad4010216d28e/xmlmemory.c#L109). I assumed that since the debug memory code used to work so far, that approach should be fine, while everything more advanced has some chance that some compiler might balk at it. But I'll be happy with some more recent and legible way to express this if I'm not held responsible for any compatibility issues that might cause…

nathan7: `offsetof` is part of C89, and I'm fairly certain that any compiler that can handle node / native node modules can also handle C89.

kkoopa:
NAN 1.8.0 is now out with this addition.

gagern:
@kkoopa thanks for the info. iojs/nan@0bc6d599f23bd00d389461ec4effa88cac7c9baf had caused me some grief, since it changes the argument of `NanObjectWrapHandle` from pointer to reference. Apart from that, everything looks good so far. Let's see what Travis thinks of this.

kkoopa:
Oh, that was unforeseen. It should have been a 1:1 replacement. I can change it to take a pointer instead.

gagern:
If you do, you break compatibility with the released 1.8.0. But you can *add* an overloaded version which accepts a pointer, and release that in 1.8.1 (hopefully *very* soon). If you do, I might revert my last commit on this branch here, and others won't experience similar breakage.

kkoopa:
It was never intended to change the argument type, just to replace the macro with a function. 1.8.0 is incorrect in this regard. Having to dereference the pointers in that way is not convenient.

kkoopa:
There, 1.8.1 is out.

defunctzombie:
let me know when this PR is using latest nan and ready for review

gagern:
@nathan7: Do you have a idea how we could detect the presence or absence of `malloc_usable_size` in a gyp environment? I like the idea a lot, but I don't feel well hardcoding its presence depending on OS or whatever, simply because I can't test on all possible platforms.

kkoopa:
You should never use `malloc_usable_size`, it is a nonstandard extension, not available everywhere, viz. Windows. It does have the same functionality in `_msize`, but that is beside the point.

It will return the amount of memory actually allocated by the `malloc` implementation (assuming that the pointer was allocated by `malloc` in the first place), not the amount of memory requested. For example, if you request 33 bytes it may actually allocate, say, 64 bytes, which is what `malloc_usable_size` would report. Nothing is specified, except that its usage will cause immense problems.

gagern:
I know `malloc_usable_size` is non-standard, which is the reason why I wanted to auto-detect its availability. The fact that the size it reports is different from the size requested shouldn't bother me: as long as all memory adjustments are based *solely* on the numbers reported this way, the result will still be sane. But if you'd rather avoid using this (for the 8 bytes overhead it saves per allocated bock), then that's fine with me, too. In that case, I think the branch is ready for review now.

kkoopa:
I don't think it cannot be reliably detected. For example, `malloc_usable_size` is available with GCC when linking against glibc, but not on Android. Then there's  Windows, then clang (which I guess is compatible with gcc in this aspect). Custom malloc implementations (Valgrind, debug builds) might cause problems as well.

Finally, it might not even save 8 bytes, if it happens to drown in the alignment padding. That is, suppose alignment is at 32 bytes, and the actual data excluding the `size_t` takes 52. 52 + 8 = 60, giving 64 bytes allocated, which would also be allocated for 52 requested bytes.

nathan7:
@gagern I don't think we can do an autoconf-style feature test for this in gyp, so… let's stick with the header approach.

gagern:
Since thread-local setup of memory handlers is difficult in #302, I've adapted my work there to this branch here as well. By now we have an asynchroneous doc reader along with a couple of test cases for it. So far everything seems to work nicely. I'd consider it ready for review.

PR #301, opened 2015-04-07
Dummy change to test Travis CI. Do not merge!

gagern:
Since Travis has trouble with node 0.8, I want to see whether current master has the same problems. So I need a merge request, but please **don't merge this!**

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -17,7 +17,7 @@
         'src/xml_syntax_error.cc',
         'src/xml_xpath_context.cc',
       ],
-      'dependencies': [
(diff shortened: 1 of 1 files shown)


gagern:
@defunctzombie: have you reported this issue somewhere?

defunctzombie:
I have not. It is also a bit unclear to be what is going on. From the logs it looks like a npm issue so possibly newer npm does work with 0.8 well? We don't install a specific version of npm. But I don't think I have seen this issue in the bcrypt repo which iirc we also test against 0.8.

PR #302, opened 2015-04-07
Make memory usage tracking thread-safe

### LINKED ISSUE #296: Async use of NanAdjustExternalMemory causes crash with Node 0.12.1

edef1c:
Fixes #296. This keeps track of allocation sizes by adding a header to
every allocation holding the size. The difference between the memory
usage we've last reported to V8 and the current memory usage is tracked
in `xml_memory_diff`, and protected by `xml_memory_mutex`. It is
frequently synchronised by `xml_memory_cb`, which is called on the main
thread through `xml_memory_handle`.

### DIFF
--- a/src/uv-compat.h
+++ b/src/uv-compat.h
@@ -2,7 +2,9 @@
 
 #if NAUV_UVVERSION < 0x10000
 
-#ifndef _WIN32
+#ifdef _WIN32
+#include <windows.h>
+#else
 #include <pthread.h>
 #endif
 
@@ -13,5 +15,15 @@ NAN_INLINE int uv_thread_equal(const uv_thread_t* t1, const uv_thread_t* t2) {
     return pthread_equal(*t1, *t2);
 #endif
 }
+
+#if !NODE_VERSION_AT_LEAST(0, 9, 2)
+NAN_INLINE uv_thread_t uv_thread_self() {
+#ifdef _WIN32
+    return GetCurrentThreadId();
+#else
+    return pthread_self();
+#endif
+}
+#endif
 
 #endif
--- a/src/uv-compat.h
+++ b/src/uv-compat.h
@@ -1,0 +1,17 @@
+#include <nan.h>
+
+#if NAUV_UVVERSION < 0x10000
+
+#ifndef _WIN32
+#include <pthread.h>
+#endif
+
+NAN_INLINE int uv_thread_equal(const uv_thread_t* t1, const uv_thread_t* t2) {
+#ifdef _WIN32
+    return *t1 == *t2;
+#else
+    return pthread_equal(*t1, *t2);
+#endif
+}
+
+#endif
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -1,6 +1,7 @@
 // Copyright 2009, Squish Tech, LLC.
 
 #include <v8.h>
+#include <uv.h>
 
 #include <libxml/xmlmemory.h>
 
@@ -16,100 +17,155 @@ namespace libxmljs {
 LibXMLJS LibXMLJS::instance;
 
 // track how much memory libxml2 is using
-int xml_memory_used = 0;
+ssize_t xml_memory_diff = 0;
+uv_async_t xml_memory_handle;
+uv_mutex_t xml_memory_mutex;
+uv_thread_t xml_thread;
 
-// wrapper for xmlMemMalloc to update v8's knowledge of memory used
-// the GC relies on this information
-void* xmlMemMallocWrap(size_t size)
-{
-    void* res = xmlMemMalloc(size);
+typedef struct alloc {
+    size_t size;
+    double data;
+} alloc_t;
 
-    // no need to udpate memory if we didn't allocate
-    if (!res)
-    {
-        return res;
-    }
+#define ALLOC_HEADER_SIZE offsetof(alloc, data)
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
-    return res;
+NAN_INLINE alloc_t* mem_data_to_header(void* ptr) {
+    return (alloc_t*) ((uintptr_t) ptr - ALLOC_HEADER_SIZE);
 }
 
-// wrapper for xmlMemFree to update v8's knowledge of memory used
-// the GC relies on this information
-void xmlMemFreeWrap(void* p)
-{
-    xmlMemFree(p);
+NAN_INLINE void* mem_header_to_data(alloc_t* ptr) {
+    return (alloc_t*) ((uintptr_t) ptr + ALLOC_HEADER_SIZE);
+}
 
-    // if v8 is no longer running, don't try to adjust memory
-    // this happens when the v8 vm is shutdown and the program is exiting
-    // our cleanup routines for libxml will be called (freeing memory)
-    // but v8 is already offline and does not need to be informed
-    // trying to adjust after shutdown will result in a fatal error
+NAUV_WORK_CB(xml_memory_cb) {
+    uv_mutex_lock(&xml_memory_mutex);
+    const ssize_t diff = xml_memory_diff;
+    xml_memory_diff = 0;
+    uv_mutex_unlock(&xml_memory_mutex);
+
+    NanAdjustExternalMemory(diff);
+}
+
+NAN_INLINE void xml_memory_update(ssize_t diff) {
+    if (diff == 0) {
+        return;
+    }
+
+    uv_thread_t current_thread = uv_thread_self();
+    if (uv_thread_equal(&xml_thread, &current_thread)) {
+        // if v8 is no longer running, don't try to adjust memory
+        // this happens when the v8 vm is shutdown and the program is exiting
+        // our cleanup routines for libxml will be called (freeing memory)
+        // but v8 is already offline and does not need to be informed
+        // trying to adjust after shutdown will result in a fatal error
 #if (NODE_MODULE_VERSION > 0x000B)
-    if (v8::Isolate::GetCurrent() == 0)
-    {
-        return;
-    }
+        if (v8::Isolate::GetCurrent() == 0)
+        {
+            return;
+        }
 #endif
-    if (v8::V8::IsDead())
-    {
-        return;
-    }
+        if (v8::V8::IsDead())
+        {
+            return;
+        }
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
+        NanAdjustExternalMemory(diff);
+    }
+    else {
+        uv_mutex_lock(&xml_memory_mutex);
+
+        xml_memory_diff += diff;
+        size_t abs_diff = diff < 0 ? -diff : diff;
+        if (abs_diff > 0x10000) {
+            uv_async_send(&xml_memory_handle);
+        }
+
+        uv_mutex_unlock(&xml_memory_mutex);
+    }
 }
 
-// wrapper for xmlMemRealloc to update v8's knowledge of memory used
-void* xmlMemReallocWrap(void* ptr, size_t size)
+// wrapper for malloc to update v8's knowledge of memory used
+// the GC relies on this information
+void* xml_malloc(size_t size)
 {
-    void* res = xmlMemRealloc(ptr, size);
+    size += ALLOC_HEADER_SIZE;
+    alloc_t* header = (alloc_t*) malloc(size);
+
+    // no need to update memory if we didn't allocate
+    if (!header)
+    {
+        return NULL;
+    }
+
+    header->size = size;
+
+    xml_memory_update(size);
+
+    return mem_header_to_data(header);
+}
+
+// wrapper for free to update v8's knowledge of memory used
+// the GC relies on this information
+void xml_free(void* ptr)
+{
+    alloc_t* header = mem_data_to_header(ptr);
+    size_t size = header->size;
+    free(header);
+
+    xml_memory_update(-size);
+}
+
+// wrapper for realloc to update v8's knowledge of memory used
+void* xml_realloc(void* ptr, size_t size)
+{
+    alloc_t* oldHeader = mem_data_to_header(ptr);
+    size_t old_size = oldHeader->size;
+
+    size += ALLOC_HEADER_SIZE;
+    alloc_t* newHeader = (alloc_t*) realloc(oldHeader, size);
 
     // if realloc fails, no need to update v8 memory state
-    if (!res)
+    if (!newHeader)
     {
-        return res;
+        return NULL;
     }
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
-    return res;
+    xml_memory_update(size - old_size);
+
+    return mem_header_to_data(newHeader);
 }
 
-// wrapper for xmlMemoryStrdupWrap to update v8's knowledge of memory used
-char* xmlMemoryStrdupWrap(const char* str)
+// replacement for strdup to update v8's knowledge of memory used
+char* xml_strdup(const char* str)
 {
-    char* res = xmlMemoryStrdup(str);
+    size_t len = strlen(str) + 1;
+    char* res = (char*) xml_malloc(len);
 
-    // if strdup fails, no need to update v8 memory state
-    if (!res)
-    {
-        return res;
+    if (!res) {
+        return NULL;
     }
 
-    const int diff = xmlMemUsed() - xml_memory_used;
-    xml_memory_used += diff;
-    NanAdjustExternalMemory(diff);
+    memcpy((void*) res, (void*) str, len);
+
     return res;
 }
 
 LibXMLJS::LibXMLJS()
 {
 
-    // populated debugMemSize (see xmlmemory.h/c) and makes the call to
-    // xmlMemUsed work, this must happen first!
-    xmlMemSetup(xmlMemFreeWrap, xmlMemMallocWrap,
-            xmlMemReallocWrap, xmlMemoryStrdupWrap);
+    uv_mutex_init(&xml_memory_mutex);
+    uv_async_init(uv_default_loop(), &xml_memory_handle, xml_memory_cb);
+    uv_unref((uv_handle_t*) &xml_memory_handle);
+    xml_thread = uv_thread_self();
+
+    // set libxml up with wrappers for memory allocation,
+    // so we can keep track of memory usage to report to V8.
+    // this needs to happen before any use of libxml.
+    xmlMemSetup(xml_free, xml_malloc,
+            xml_realloc, xml_strdup);
 
     // initialize libxml
     LIBXML_TEST_VERSION;
-
-    // initial memory usage
-    xml_memory_used = xmlMemUsed();
 }
 
 LibXMLJS::~LibXMLJS()
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -2,6 +2,7 @@
 
 #include <v8.h>
 #include <uv.h>
+#include "uv-compat.h"
 
 #include <libxml/xmlmemory.h>
 
--- a/vendor/libxml/include/libxml/xmlversion.h
+++ b/vendor/libxml/include/libxml/xmlversion.h
@@ -90,11 +90,20 @@ XMLPUBFUN void XMLCALL xmlCheckVersion(int version);
  *
  * Whether the thread support is configured in
  */
-#if 0
+#if 1
 #if defined(_REENTRANT) || defined(__MT__) || \
     (defined(_POSIX_C_SOURCE) && (_POSIX_C_SOURCE - 0 >= 199506L))
 #define LIBXML_THREAD_ENABLED
 #endif
+#endif
+
+/**
+ * LIBXML_THREAD_ALLOC_ENABLED:
+ *
+ * Whether the allocation hooks are per-thread
+ */
+#if 0
+#define LIBXML_THREAD_ALLOC_ENABLED
 #endif
 
 /**


gagern:
Wrt code style, I prefer the inline functions with semantic names to encapsulate the raw pointer arithmetic. You might consider that a matter of taste, but if you aim for understandable code, I think names for these computations would help. I know that on some very exotic architectures the conversion between pointer and integer might be cause for concern as well, but I doubt libxml would compile on any such architecture in any case.

The `uv_unref` nicely avoids the problem of the stale async. Hadn't known about that before, and completely missed it while reading nathan7/libxmljs@2b0224a7ad4e845b1145f01c67a83932d580578c.

There is one thing which causes great concern for me: you are using the `async` for *all* memory adjustment reports, even on the main thread. While I have verified that it is indeed possible to send an async from the event dispatch thread, the handler will only be run once control returns to the main loop. So if you have some highly synchroneous work load, i.e. some script which spends hours composing complicated XML documents without ever returning control back to the event loop, then V8 won't be notifier of memory changes in all that time. I assume this might be a problem.

If there is a reason why this is not a problem, then I wonder whether we can drop the `v8::Isolate::GetCurrent() == 0` and `v8::V8::IsDead()` case distinctions. If someone is alive to execute an async sent to the main event loop, then obviously the engine is not dead yet.

You should probably cherry-pick gagern/libxmljs@43d035271babb90039f3c0760be57db3ec1aec07 into your code, or simply reproduce its effect. After all, having thread-safe memory handling is one aspect, but making the rest of libxml thread-safe as well is important for #296, too.

nathan7:
> I know that on some very exotic architectures the conversion between pointer and integer might be cause for concern as well, but I doubt libxml would compile on any such architecture in any case.

`uintptr_t` is guaranteed to be the same size as a pointer, and guaranteed to be valid to cast pointers to and from.

> There is one thing which causes great concern for me: you are using the async for all memory adjustment reports, even on the main thread. While I have verified that it is indeed possible to send an async from the event dispatch thread, the handler will only be run once control returns to the main loop. So if you have some highly synchronous work load, i.e. some script which spends hours composing complicated XML documents without ever returning control back to the event loop, then V8 won't be notifier of memory changes in all that time. I assume this might be a problem.

Acknowledged — I had some test code that did take care of this, but it used pthreads directly, which wouldn't go over well on Windows.

> If there is a reason why this is not a problem, then I wonder whether we can drop the v8::Isolate::GetCurrent() == 0 and v8::V8::IsDead() case distinctions. If someone is alive to execute an async sent to the main event loop, then obviously the engine is not dead yet.

Yep, I can't actually get that case to trigger. Tests run just fine with this, both libxmljs and node-libxslt:
```diff
diff --git a/src/libxmljs.cc b/src/libxmljs.cc
index c7a5d8a..a91eab1 100644
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -37,12 +37,12 @@ NAUV_WORK_CB(xml_memory_cb) {
 #if (NODE_MODULE_VERSION > 0x000B)
     if (v8::Isolate::GetCurrent() == 0)
     {
-        return;
+        abort();
     }
 #endif
     if (v8::V8::IsDead())
     {
-        return;
+        abort();
     }
 
     uv_mutex_lock(&xml_memory_mutex);
```

> The uv_unref nicely avoids the problem of the stale async. Hadn't known about that before, and completely missed it while reading nathan7/libxmljs@2b0224a.

Thanks!

> You should probably cherry-pick gagern/libxmljs@43d0352 into your code, or simply reproduce its effect. After all, having thread-safe memory handling is one aspect, but making the rest of libxml thread-safe as well is important for #296, too.
Yep, I'll do that, though I'm somewhat wary of patching our vendored lib.

nathan7:
I've made some more progress, so, status report!

> > Wrt code style, I prefer the inline functions with semantic names to encapsulate the raw pointer arithmetic. You might consider that a matter of taste, but if you aim for understandable code, I think names for these computations would help.

> Concern noted, and I'll take a look at doing that.

I've patched this up now, the code does look a *lot* more readable now, thank you!

> > You should probably cherry-pick gagern/libxmljs@43d0352 into your code, or simply reproduce its effect. After all, having thread-safe memory handling is one aspect, but making the rest of libxml thread-safe as well is important for #296, too.

> Yep, I'll do that, though I'm somewhat wary of patching our vendored lib.

Done!

> If there is a reason why this is not a problem, then I wonder whether we can drop the v8::Isolate::GetCurrent() == 0 and v8::V8::IsDead() case distinctions. If someone is alive to execute an async sent to the main event loop, then obviously the engine is not dead yet.

I'm still somewhat wary of taking that code out, even though it appears to do nothing. In particular, when I fix my patches to be aware of the synchronous case, this code might run while node is shutting down. It's harmless in the meanwhile.

> > There is one thing which causes great concern for me: you are using the async for all memory adjustment reports, even on the main thread. While I have verified that it is indeed possible to send an async from the event dispatch thread, the handler will only be run once control returns to the main loop. So if you have some highly synchronous work load, i.e. some script which spends hours composing complicated XML documents without ever returning control back to the event loop, then V8 won't be notifier of memory changes in all that time. I assume this might be a problem.

> Acknowledged — I had some test code that did take care of this, but it used pthreads directly, which wouldn't go over well on Windows. Advice is definitely welcome on that.
> I was considering having a thread-local "are we in V8" variable, but I'm not sure that's worth the complexity.

This remains a TODO, and I'm working on figuring out how to do this in the nicest way possible with the tools libuv gives us.

gagern:
To special-case the V8 thread: you can use [`uv_thread_self`](http://docs.libuv.org/en/v1.x/threading.html#c.uv_thread_self) when initializing the library to store a reference to the main thread. Then later on you can compare that against the current thread to see whether you are on the V8 thread. Note that in older versions of libuv, thta function returns some integer type.

nathan7:
> To special-case the V8 thread: you can use uv_thread_self when initializing the library to store a reference to the main thread. Then later on you can compare that against the current thread to see whether you are on the V8 thread.

Done!

> Note that in older versions of libuv, that function returns some integer type.

While `pthread_t` and `HANDLE` (on Windows) are both integer types, this appears to be wrapped in `uv_thread_t` down to the version of libuv that node 0.8 uses.

gagern:
And `uv_thread_equal` was introduced in libuv/libuv@59658a8de7cc05a58327a164fd2ed4b050f8b4f4 and is contained in libuv versions starting from 1.0.0. Before that `uv_thread_self` used to return an `unsigned long` which it obtained by a brute force cast. You will notice that Travis CI fails on 0.10 as well for the current state of affairs here.

gagern:
Personally I find it hard to review a sequence of changes if they are all squashed into a single commit again and again. It means I can't concentrate on what changed since the last iteration. But that's a matter of taste, I guess.

nathan7:
@gagern 
> Personally I find it hard to review a sequence of changes if they are all squashed into a single commit again and again. It means I can't concentrate on what changed since the last iteration. But that's a matter of taste, I guess.

If you want to review this as individual commits, check out https://github.com/polotek/libxmljs/compare/master...nathan7:memory-tracking-wip.
I try to keep my pull requests limited to semantic commits, but you I always keep the full history on the `-wip` branch. Sorry for not making that clear before.

nathan7:
> > If there is a reason why this is not a problem, then I wonder whether we can drop the v8::Isolate::GetCurrent() == 0 and v8::V8::IsDead() case distinctions. If someone is alive to execute an async sent to the main event loop, then obviously the engine is not dead yet.

> I'm still somewhat wary of taking that code out, even though it appears to do nothing. In particular, when I fix my patches to be aware of the synchronous case, this code might run while node is shutting down. It's harmless in the meanwhile.

This code appears to be necessary in the synchronous case, so I'll get on that:
```
OK: 323 assertions (1792ms)
[Thread 0x7ffff4be0700 (LWP 15310) exited]
[Thread 0x7ffff53e1700 (LWP 15309) exited]
[Thread 0x7ffff5be2700 (LWP 15308) exited]
[Thread 0x7ffff63e3700 (LWP 15307) exited]
[Thread 0x7ffff6be4700 (LWP 15306) exited]

==== C stack trace ===============================

 1: V8_Fatal
 2: v8::internal::Isolate::Current()
 3: v8::Isolate::GetCurrent()
 4: NanAdjustExternalMemory(int)
 5: libxmljs::xml_memory_update(long)
 6: libxmljs::xmlMemFreeWrap(void*)
 7: xmlCleanupCharEncodingHandlers
 8: xmlCleanupParser
 9: libxmljs::LibXMLJS::~LibXMLJS()
10: ??
11: ??
12: __libc_start_main
13: _start

Program received signal SIGILL, Illegal instruction.
v8::base::OS::Abort () at ../../deps/v8/src/base/platform/platform-posix.cc:212
212	    V8_IMMEDIATE_CRASH();
=> 0x000000000126d41f <v8::base::OS::Abort()+15>:	0f 0b	ud2
```

nathan7:
@gagern 
I've [moved the V8 aliveness checks onto the synchronous case](https://github.com/polotek/libxmljs/commit/4a43c03a9ba9075635a21d27675a565d3865efad), and [renamed and documented the new allocation approach](https://github.com/polotek/libxmljs/commit/591ae6770ea9ca6f184c923a06adb61d8528ae2f).
I think we're pretty much there, but as usual, a second brain and pair of eyes is greatly appreciated :heart:

nathan7:
> > Note that in older versions of libuv, that function returns some integer type.

> While pthread_t and HANDLE (on Windows) are both integer types, this appears to be wrapped in uv_thread_t down to the version of libuv that node 0.8 uses. I'm not sure what issue I should be seeing here. I'm using uv_thread_equal(const uv_thread_t*, const uv_thread_t*) for my comparison.

It seems this one has come back to bite me, breaking compilation on node 0.10 (`UV_VERSION_MAJOR == 0`)

nathan7:
@gagern
I've [fixed the lack of `uv_thread_equal` on pre-1.0 libuv](https://github.com/nathan7/libxmljs/commit/959ca52122afe6e8565fa01f7ee3a32f3a99c78f) using your approach.

nathan7:
It appears I need to do some more libuv compatibility work, because `uv_thread_self` apparently didn't exist in node 0.8.
```
../src/libxmljs.cc: In function ‘void libxmljs::xml_memory_update(ssize_t)’:
../src/libxmljs.cc:55:49: error: ‘uv_thread_self’ was not declared in this scope
     uv_thread_t current_thread = uv_thread_self();
                                                 ^
../src/libxmljs.cc: In constructor ‘libxmljs::LibXMLJS::LibXMLJS()’:
../src/libxmljs.cc:159:33: error: ‘uv_thread_self’ was not declared in this scope
     xml_thread = uv_thread_self();
                                 ^
```

nathan7:
While `uv_thread_self` appears to have started existing in node 0.8.10, it appears to [be broken on pre-1.0 libuv](https://github.com/libuv/libuv/compare/8a8cff4b34c6dbaf6d31a2c9dda8afa3b1509cf0...6591d000d13714c9e362942fbeaeca2b29507a25) (it doesn't actually work with `uv_thread_equals`).
That's a pickle.

nathan7:
It appears that as long as we're comparing return values from `uv_thread_self`, we're in the clear.
The issue only shows up when you compare a "real" `uv_thread_t` (such as the ones returned by `uv_thread_create`) to one returned from `uv_thread_self`.

I think that's fine, we don't do that, and it only shows up on a rather narrow platform / node version combo.
The worst that could happen if we did that is that our memory usage is accidentally reported asynchronously in the synchronous case. But we don't do that.

Just a comment noting this would do here, really.

gagern:
Looks good. The type `uv_thread_t` for `xml_thread` has me somewhat worried, since `uv_thread_self` will sometimes return a `long` instead which might make compilers balk. Not sure which versions are affected. Apparently not the current version of either node 0.10 or 0.12, but probably some past versions. I wonder whether you should be including a replacement for `uv_thread_self` as well in libuv pre 1.0. And of course that would be a candidate for nauv in nan as well.

Your header doesn't really have to include `<nan.h>` but doing so should cause no harm either. I'd really like to do a performance test soon, but apart from that things look good.

nathan7:
I've [implemented `uv_thread_self` for pre-0.9.2 node versions](https://github.com/nathan7/libxmljs/commit/da75bcab818ecaf77e8acde13dcf7e6bc7072fb8), using the pre-1.0 behaviour. I'm not sure about the `unsigned long` matter — there's a node commit by @tjfontaine reverting the libuv change that switched the return type, so that might be worth checking on.

nathan7:
If someone on Windows, with node 0.8, could verify that this doesn't crash the libxmljs tests, and *does* crash the node-libxslt tests, that'd be splendid:
```diff
diff --git a/src/libxmljs.cc b/src/libxmljs.cc
index f306c24..3b3e136 100644
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -73,6 +73,7 @@ NAN_INLINE void xml_memory_update(ssize_t diff) {
         NanAdjustExternalMemory(diff);
     }
     else {
+        abort();
         uv_mutex_lock(&xml_memory_mutex);
 
         xml_memory_diff += diff;
```

nathan7:
As far as I can tell, `uv_thread_t` is always `unsigned long` or larger to match the OS. I think we can always safely upcast OS thread IDs. I'd love to test this across every node version since 0.8.0, on every OS, but I can really only do that for Linux and OS X currently.

nathan7:
I've kicked off what is [probably the most insane single build in Travis CI history](https://travis-ci.org/nathan7/libxmljs/builds/57626939), but it seems it isn't intending to build it on OS X as requested.
I'll be patiently waiting for that to complete — if it isn't done by the end of today, I'll probably kick off a bunch of builds of my own.

nathan7:
Results of the giant Travis test, across 102 node versions:

 * 0.8.13: Ships with an npm so broken even `npm test` doesn't work, but, rest assured, that actually does work locally if I run it by hand.
 * 0.9.{9..12}, 0.10.0: Mysteriously fails due to `node::MakeCallback(v8::Handle<v8::Object>, v8::Handle<v8::Function>, int, v8::Handle<v8::Value>*)` being AWOL, not our problem.
 * 0.11.{0..12}: Everything breaks, C++ errors of Wagnerian fiercety are produced by Nan. [That's not my department](https://youtu.be/TjDEsGZLbio?t=1m15s).
 * 0.11.14: Yep, that's my fault: `../src/libxmljs.cc:56:53: error: ‘uv_thread_equal’ was not declared in this scope`
 * iojs: Whoops. The node-gyp I bundled doesn't deal with iojs correctly. I can assure you that everything works just fine with the right node-gyp, however.

nathan7:
Meanwhile, my Mac Mini is busily installing the same set of node versions. I'll report back on that once testing kicks off.

gagern:
I'm currently cooking up an asynchroneous version of `Document.fromXml` and reading a 30+M file on my system in parallel to test performance. At least that's the idea; right now I'm doing this on my code and the TLS variable mysteriously changes its value. So either I've got some memory corruption somewhere, or a bug in libuv or pthreads or so. I consider the former far more likely. Investigating…

gagern:
OK, I believe we do have a performance problem with this approach. The biggest XML file on my system is `/usr/share/foomatic/db/source/opt/gutenprint-ijs.5.2-pagesize.xml` with 34M. I parsed 20 times in parallel:

```js
var testInput = fs.readFileSync(testFile);
for (var i = 0; i < 20; ++i)
  libxmljs.Document.fromXmlAsync(testInput, function(err, doc) {
    if(err) throw err; console.log("Got document."); });
```

The result:

         my #300   your #302
    real 0m23.811s 0m24.658s
    user 0m24.483s 0m40.894s
    sys  0m04.372s 0m16.335s

The #300 test case is in [asyncTest1](https://github.com/gagern/libxmljs/tree/asyncTest1), the one for #302 in [asyncTest2](https://github.com/gagern/libxmljs/tree/asyncTest2). So while the wall clock time appears to be the both for both, your branch will spend *a lot* more CPU resources, in user as well as in system space. I still prefer #300. Have I convinced you?

gagern:
The way I read this, libxml seems to essentially serialize access to the parser, so there is only ever one thread accessing the parser, which is the bottleneck in my use case. But with your code, the event dispatch thread is almost as busy as that worker, while in mine it appears to be mostly idle, and could do other work. Of course, it is possible that if the V8 thread had something to do, your async events would simply coalesce, and therefore have almost no effect on the performance of the main thread either. I wonder how to test this theory.

nathan7:
Those results are indeed quite damning. I'll take a look later today, and see if I can work on that.
If I can be convinced that your approach is better, all the better! Less work for me [=

src/uv-compat.h
```diff
@@ -0,0 +1,29 @@
...
+
+NAN_INLINE int uv_thread_equal(const uv_thread_t* t1, const uv_thread_t* t2) {
+#ifdef _WIN32
+    return *t1 == *t2;
+#else
+    return pthread_equal(*t1, *t2);
+#endif
+}
+
+#if !NODE_VERSION_AT_LEAST(0, 9, 2)
```
saghul: See https://github.com/libuv/libuv/commit/6591d000d13714c9e362942fbeaeca2b29507a25 and https://github.com/libuv/libuv/commit/a31024865b3146597cecc3d7f5bdf251da8b1a9b

nathan7: The thread's HANDLE will do for us — as long as uv_thread_self() returns some form of thread ID.

nathan7:
I've [modified the tracking a little, to prevent overly frequent synchronisation](https://github.com/nathan7/libxmljs/commit/e5064ddbd77f14eec7a67f7b1cb38d4fe9c01dd3): external memory is allowed to diverge up to 64k.
That might be enough to solve the CPU usage issues.
I'm also looking into hooking [`V8::AddGCPrologueCallback`](http://izs.me/v8-docs/classv8_1_1V8.html#a49c016f17c67f700387f801b2b29b5ab) (accessible in Nan through [`NanAddGCPrologueCallback`](https://github.com/rvagg/nan#nanaddgcprologuecallbackgcprologuecallback-callback-gctype-gc_type_filterkgctypeall)), though I'm not entirely sure that's useful at all.

nathan7:
Nope, that was pointless.
```
      #300     new #302  old #302
real  12.672s  18.229s   20.901s
user  27.730s  39.243s   45.397s
sys    5.827s  17.710s   22.030s
```

defunctzombie:
@nathan7 @gagern Ok, so after all this research and attempts, what is the patch I should be reviewing and looking to merge?

gagern:
@defunctzombie If I understand things correctly, nathan is primarily concerned with the complexity of #300. I'd say please have a look at both, let us know whether you share that opinion, and how much performance you'd be willing to sacrifice in order to avoid it.

gagern:
I doubt adding a gc prologue callback will be of any use: if it gets executed, the system has already decided to run a gc sweep, which as far as I understand things is the very same decision for which it requires memory adjustments.

nathan7:
@defunctzombie I'm concerned with #300 being more complex and requiring cooperation from the embedder, which is a breaking change at the C++ interface but not the JS interface.
I'm not sure I am qualified to review that code. You should probably find someone more well-versed in C++.

gagern:
Well, up till now libxmljs made no attempt to support worker threads. Any third party code which used it in this fashion, like node-libxslt did, did so either out of ignorance or knowingly hoping for things not to break. The way I see it, this is the time you are first offering multi-threading support at all, and it's up to you to set down the guidelines for how you do it.

It's only after you start releasing this feature that you should stick to the API once chosen. Which is of course easier if you from the beginning require implementers to use sentinels or whatever, than if you guarantee that they won't have to worry about anything.

defunctzombie:
@gagern @nathan7 I took a look at both patches and I personally am leaning toward the one where the user doesn't have to do anything special. I think the Sentinel stuff in patch #300 could lead to more issues and opens up API surfaces for us to maintain.

For this patch, is the single thread perf much worse? Can we maybe detect single thread use and avoid dropping perf unless we know user needs the multi-thread stuff?

This patch also needs tests for this so that it is not broken in future updates.

defunctzombie:
I'll also add that I like the `features` thing from patch #300 and maybe that could be a separate PR? Dunno if anyone actually cares about them but it seemed potentially useful.

gagern:
> I took a look at both patches and I personally am leaning toward the one where the user doesn't have to do anything special. I think the Sentinel stuff in patch #300 could lead to more issues and opens up API surfaces for us to maintain.

OK, then let's aim for that.

> For this patch, is the single thread perf much worse? Can we maybe detect single thread use and avoid dropping perf unless we know user needs the multi-thread stuff?

The `uv_thread_equal(&xml_thread, &current_thread)` case should do just that, as far as our own memory handling code is concerned. I fear that the biggest performance impact would likely come from libxml itself being compiled with threading support, since that means locks in many places. But that effect is independent of which approach we use to tie in memory management, and hopefully not enough to warrant compiling the library in two different versions. I haven't done any benchmarks yet.

> This patch also needs tests for this so that it is not broken in future updates.

We could build on [asyncTest2](https://github.com/gagern/libxmljs/commit/ed4b8ca31446d3ece6faf994e0543946e48573fe) for that. Either turn it into a compiled-code test case or add some proper error-storing code (without using v8 data structures) to make `fromXmlAsync` a proper library addition. I'd prefer the latter.

> I'll also add that I like the features thing from patch #300 and maybe that could be a separate PR? Dunno if anyone actually cares about them but it seemed potentially useful.

Yes, I can extract that part of #300 and file it separately.

gagern:
In [my version](https://github.com/gagern/libxmljs/commits/memory-tracking) of Nathan's memory-tracking branch, I included the list of features and implemented an asynchroneous parser including error collecting. It doesn't work yet, there seems to be some memory trouble somewhere, but there is an even more severe problem: up until now, we had not set `HAVE_PTHREAD_H` nor `HAVE_WIN32_THREADS`, thus essentially disabling threading support again. With that define added in (in gagern/libxmljs@fc66a1cc21e9d4bfab34ea712e94a10768566d8d) things changed dramatically: now *many* things are suddenly thread-local, see [the global state object](https://github.com/polotek/libxmljs/blob/63a4cf205a221f74471d47dc8f0e7dc7ad247bfc/vendor/libxml/include/libxml/globals.h#L116-164) for comparison. In particular, the memory functions are now thread-local! This means that we have to call `xmlMemSetup` in every worker thread! With the approach from #300 that would be an easy thing to do: simply add some more calls to the constructor of the `WorkerSentinel` class. But I don't see how we can achieve this without cooperation from the calling code which set up the worker thread. So personally I'd see this as a strong indication that we should switch to #300 instead of this here, but if you can think of another approach, I'd be glad to hear that.

defunctzombie:
What happens if we just remove the code that tells v8 about the amount of externally allocated memory? Also, what is it that is starting worker threads? I am going through the original issue and trying to figure out why this is even a thing in the first place. Are worker threads being launched by some other lib? Are these worker threads uv workers or some other thread?

My main concern was with adding API around this when it seems to be a very niche and scoped use case? Maybe that is not the case but it certainly seems that way.

Can we just use global xmlMemSetup and memory tracking? What is the issue with keeping memory tracking global vs per thread?

gagern:
> What happens if we just remove the code that tells v8 about the amount of externally allocated memory?

We could end up with a bunch of stale documents which don't get garbage-collected because v8 doesn't know it's eating tons of memory.

> Also, what is it that is starting worker threads? I am going through the original issue and trying to figure out why this is even a thing in the first place. Are worker threads being launched by some other lib? Are these worker threads uv workers or some other thread?

albanm/node-libxslt#9 was the starting point. node-libxslt depends on libxmljs to provide libxml bindings, but performs its operations asynchroneously. It's using the uv-based worker implementation from nan.

There might be other libraries doing something similar without my knowledge. After all, there is quite a number of [packages depending on libxmljs](https://www.npmjs.com/browse/depended/libxmljs), and I haven't had a closer look at what all of these do internally.

> My main concern was with adding API around this when it seems to be a very niche and scoped use case? Maybe that is not the case but it certainly seems that way.

It might be a niche use case now, I don't know. It might become more common in the future, as more wrappers emergw around libs that depend on libxml2. The node way to do expensive things is asynchroneously, the way I understand it.

The way I see it, you have to decide whether you want to officially support multi-threading. If you do, I believe you need some API around this, no matter how uncommon the use case is at the moment. You won't have to worry about that new API requirement breaking any existing code, since with node 0.12 the current approach is broken in any case, and as long as you ensure that not using the new API breaks loudly enough, that should be fine. If it's only node-libxslt which is affected, then I can gladly take care of getting that adapted.

Or you decide not to support multi-threading, which means node-libxslt will either have to drop (or fake) its asynchroneous operations, or change its dependency to something which supports multi-threading. Seeing as I'm reasonably happy with #300, I'd suggest forking libxmljs to provide that functionality for node-libxslt. So the question is this: do you want a fork of this project to deal with the threading issues, or would you rather avoid the confusion such a fork brings with it?

> Can we just use global xmlMemSetup and memory tracking? What is the issue with keeping memory tracking global vs per thread?

Looking at the `configure.ac` of libxml2, it seems as if `WITH_THREADS` can only get set if either `HAVE_WIN32_THREADS` or `HAVE_PTHREAD_H` gets set at the same time. So the current state of the `config.h` in this branch here, with `WITH_THREADS` set but neither of the other two, can't happen in a properly autotooled build, and is therefore unsupported. It won't result in reentrant code, since it won't have access to synchronization primitives like mutexes.

PR #306, opened 2015-04-17
Add xinclude support

hildjj:
The option processing was already there for it, but it wasn't turned on.

Note: this *does* enable nanohttp, which is potentially unwanted; we might prefer to use node's HTTP stack.  However, that would likely mean making an async version of parseXml, which seemed like a larger change than I wanted to make without discussing it with you first.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -19,7 +19,8 @@
       ],
       'dependencies': [
         './vendor/libxml/libxml.gyp:libxml'
-      ]
+      ],
+      'defines': ['LIBXML_XINCLUDE_ENABLED']
     }
   ]
 }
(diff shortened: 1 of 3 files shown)


defunctzombie:
Do not think this belongs here personally. What does xinclude do? The scope of this module is to do CPU only xml processing tasks.

hildjj:
I think the scope of this module should be to provide access to the pieces of libxml2 that people want.  The XInclude spec is here: http://www.w3.org/TR/xinclude/.  It allows pulling in parts of your document from other places.  XInclude is particularly handy when writing RFCs using the xml2rfc tooling; it allows you to keep your examples, artwork, etc, in files that can be edited more easily in separate files.  Thing SVG, for example.

defunctzombie:
Please see this long discussion about module scope here: https://github.com/polotek/libxmljs/issues/173

Possible to revisit but honestly not sure how much time I can devote to that. Also, this PR has no tests so adding such a feature with no actual testing won't be merged.

hildjj:
I would have been happy to add tests. I will re-invent the wheel in plain Javascript, though.

PR #311, opened 2015-05-10
Fixes build on iojs 2.0.0 with V8 4.2.x

marcominetti:
Seems updating nan fixes the problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "bindings": "1.2.1",
-    "nan": "1.7.0"
+    "nan": "1.8.4"
   },
   "devDependencies": {
     "nodeunit": "0.9.0"


package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "bindings": "1.2.1",
-    "nan": "1.7.0"
+    "nan": "^1.8.0"
```
defunctzombie: please follow existing dependency pattern

arthurschreiber:
Awesome!

defunctzombie:
@marcominetti isn't there a newer version of nan?

package.json
```diff
@@ -5,7 +5,7 @@
     "Jeff Smick"
   ],
   "description": "libxml bindings for v8 javascript engine",
-  "version": "0.14.0",
+  "version": "0.14.1",
```
defunctzombie: Tip: don't do this in PRs. It will never be merged because it is up to the maintainer to decide what the next version will be.

marcominetti: Yep, I did it to prioritize download from my private npm...

defunctzombie:
handled manually and released v0.14.1

PR #316, opened 2015-06-06
Bindings for list of features from xmlHasFeature

gagern:
As @defunctzombie [requested](https://github.com/polotek/libxmljs/pull/302#issuecomment-106978992), I've extracted the feature-list from #300 so we can get this included independently. I'm not adding any tests to this, since I expect that we will be using this in other tests for threading pretty soon.


gagern:
Since I've recently found new problems with #302, it might well be that we will eventually want to merge #300 instead, which already contains what this commit here is doing. So if #300 gets merged *soon*, then this here becomes redundant. If it takes longer, having this in first won't hurt much, since resolving the conflicts between the two should be straight-forward.

PR #317, opened 2015-07-01
New Document method to read XML from file

sergeche:
Added `XmlDocument::FromXmlFile` method to read XML from file.

The reason behind this method is to use parsed XML document which contains references to other files. For example, to parse XML document as XSLT stylesheet with https://github.com/albanm/node-libxslt

Without proper file-based xml parsing context it is not possible to parse XSLT stylesheets that contain, for example, `<xsl:import href="..."/>` references.

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,6 +9,9 @@ var Document = require('./lib/document');
 /// parse an xml string and return a Document
 module.exports.parseXml = Document.fromXml;
 
+/// parse an xml file and return a Document
+module.exports.parseXmlFile = Document.fromXmlFile;
+
 /// parse an html string and return a Document
 module.exports.parseHtml = Document.fromHtml;
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -146,3 +146,10 @@ module.exports.fromXml = function(string, options) {
     return bindings.fromXml(string, options || {});
 };
 
+/// parse given file contents into a xml document
+/// @param string file path to file
+/// @return a Document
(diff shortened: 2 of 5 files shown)


hildjj:
To make this parallel with FromHTML, you might instead add a "baseUrl" and "encoding" option, which you would use to set the 3rd and 4th parameters to xmlReadMemory.

sergeche:
I’ve tried with `baseUrl` but it didn’t work. Looks like it’s not about document’s URI but parser context.

defunctzombie:
Hm. I originally removed this because node already has file reading support but I can see how the links feature could be useful.

Does the test actually have links in the file and tests that the desired functionality works and that we are not just replication fs.read ?

sergeche:
@defunctzombie no, there’s no such tests yet because I’ve used it with libxslt. But I think it might work with X:Include, will figure it out

Klortho:
Regarding baseUrl, I just made this [pull request](https://github.com/polotek/libxmljs/pull/388) to add it to FromXml.

FromXmlFile would be nice to have too, but if you add it, I'd encourage you not to copy-paste-hack so much code. This was the reason baseUrl was in FromHtml and not FromXml in the first place.

rchipka:
@sergeche What's the advantage of using this method vs reading the file in JS?

sergeche:
It’s for XInclude and for XSL imports where files are read by libxml and require proper file base context for relative file paths

PR #318, opened 2015-07-10
Get Travis to build for iojs as well

gagern:
Do you want Travis to build against io.js?

The problem with `node-gyp` not working with iojs can be solved by using the version shipped with iojs, not the version from the latest npm we install separately.  See https://github.com/nodejs/io.js/issues/433 for details. So we update npm to latest upstream only for old versions of node.js, not for the current version or for iojs.


defunctzombie:
:+1:

PR #323, opened 2015-07-22
Segfault prevention/safety

rc0x03:
This pull request causes XmlNodes to Ref() their parents. This causes the parent node's ref count to increase, which means that it won't be GC'ed until the ref count goes back down to 0. When the child node's destructor is called, it Unref()'s the parent node which decreases the parent ref count. This forces parent nodes to stick around while there are still children that haven't been freed.

This pull request also causes libxmljs to check that `xml_obj` hasn't been freed before it tries to access any of its properties. This mitigates segfaults caused by the XmlNode destructor accessing `xml_obj` properties. Though the parent Ref() part of this pull request already fixes most known segfault issues, this is still useful as an extra safety check to provide protection against undiscovered segfaults.

This pull request passes all tests, including the `double_free` test, without any known issues or memory leaks.

*Note:* Both of these features are almost identical to the memory management steps Ruby has taken in [libxml-ruby](https://github.com/xml4r/libxml-ruby/blob/master/ext/libxml/ruby_xml_node.c).

### DIFF
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -97,8 +97,26 @@ char* xmlMemoryStrdupWrap(const char* str)
     return res;
 }
 
+// callback function for `xmlDeregisterNodeDefault`
+void xmlDeregisterNodeCallback(xmlNode* xml_obj)
+{
+    if (xml_obj->_private)
+    {
+        XmlNode* node = static_cast<XmlNode*>(xml_obj->_private);
+
+        // flag the XmlNode object as freed
+        node->freed = true;
+
+        // save a reference to the doc so we can still `unref` it
+        node->doc = xml_obj->doc;
+    }
+    return;
+}
+
 LibXMLJS::LibXMLJS()
 {
+    // set the callback for when a node is about to be freed
+    xmlDeregisterNodeDefault(xmlDeregisterNodeCallback);
 
     // populated debugMemSize (see xmlmemory.h/c) and makes the call to
     // xmlMemUsed work, this must happen first!
--- a/src/xml_node.cc
+++ b/src/xml_node.cc
@@ -179,17 +179,45 @@ XmlNode::New(xmlNode* node)
 }
 
 XmlNode::XmlNode(xmlNode* node) : xml_obj(node) {
+    this->freed = false;
     xml_obj->_private = this;
 
     // this will prevent the document from being cleaned up
     // we keep the document if any of the nodes attached to it are still alive
     XmlDocument* doc = static_cast<XmlDocument*>(xml_obj->doc->_private);
     doc->ref();
+
+    // check that there's a parent node with a _private pointer
+    // also check that the parent node isn't the doc since we already Ref() the doc once
+    if (xml_obj->parent != NULL &&
+        xml_obj->parent->_private != NULL &&
+        (void*)xml_obj->doc != (void*)xml_obj->parent)
+    {
+        static_cast<XmlNode*>(xml_obj->parent->_private)->Ref();
+    }
 }
 
 XmlNode::~XmlNode() {
-    xml_obj->_private = NULL;
 
+    // check if `xml_obj` has been freed so we don't access bad memory
+    if (this->freed)
+    {
+
+        // unref the doc using the doc reference we saved in the `flagNode` callback
+        if (this->doc)
+        {
+            XmlDocument* doc = static_cast<XmlDocument*>(this->doc->_private);
+            doc->unref();
+        }
+
+        // set doc to null for good measure?
+        // this->doc = NULL;
+
+        // return so we don't attempt to use `xml_obj`
+        return;
+    }
+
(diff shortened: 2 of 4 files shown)


src/libxmljs.cc
```diff
@@ -97,8 +97,24 @@ char* xmlMemoryStrdupWrap(const char* str)
     return res;
 }
 
+// callback function for `xmlDeregisterNodeDefault`
+void flagNode(xmlNode* xml_obj) {
+  if (xml_obj->_private) {
```
defunctzombie: follow existing code style

src/xml_node.h
```diff
@@ -12,6 +12,12 @@ class XmlNode : public node::ObjectWrap {
 
     xmlNode* xml_obj;
 
+    // boolean value to check if `xml_obj` was already freed
+    bool isFree;
```
defunctzombie: follow existing code style

rc0x03:
Do you have any specific advice on following the existing code style?

defunctzombie:
Look at the existing code in the file :) Specifically spaces and variable naming conventions.

rc0x03:
@defunctzombie How's that?

test/ref_integrity.js
```diff
@@ -25,3 +25,26 @@ module.exports.references = function(assert) {
...
 };
+
+// test that double-freeing XmlNode's doesn't cause a segfault
+module.exports.double_free = function(assert) {
+    var children = null;
+
+    // stick this portion of code into a self-executing function so
+    // its internal variables can be garbage collected
+    (function(){
+      var html = '<html><body><div><span></span></div></body></html>';
```
defunctzombie: 4 spaces

src/libxmljs.cc
```diff
@@ -97,8 +97,26 @@ char* xmlMemoryStrdupWrap(const char* str)
     return res;
 }
 
+// callback function for `xmlDeregisterNodeDefault`
+void xmlDeregisterNodeCallback(xmlNode* xml_obj)
+{
+  if (xml_obj->_private)
```
defunctzombie: 4 spaces

src/xml_node.cc
```diff
@@ -179,17 +179,45 @@ XmlNode::New(xmlNode* node)
...
-    xml_obj->_private = NULL;
 
+    // check if `xml_obj` has been freed so we don't access bad memory
+    if (this->freed)
+    {
+
+        // unref the doc using the doc reference we saved in the `flagNode` callback
+        if (this->doc)
+        {
+          XmlDocument* doc = static_cast<XmlDocument*>(this->doc->_private);
```
defunctzombie: 4 spaces

src/xml_node.cc
```diff
@@ -198,6 +226,20 @@ XmlNode::~XmlNode() {
     // It will be freed when the doc is freed
     if (xml_obj->parent == NULL)
       xmlFreeNode(xml_obj);
+
+    // if there's a parent then Unref() it
+    else if (xml_obj->parent->_private != NULL &&
+            (void*)xml_obj->doc != (void*)xml_obj->parent)
+    {
+
+            XmlNode* parent = static_cast<XmlNode*>(xml_obj->parent->_private);
```
defunctzombie: 4 spaces

rc0x03:
@defunctzombie Thanks for pointing that out. I believe that's all the spacing issues. Any other issues? Is there anything wrong with this solution?

rc0x03:
I ran a simple test parsing 1,000 large and unique HTML documents using node-osmosis to compare the current master with this pull request. Both tests were steady at around 230Mb of RAM without any segfaults.

I added unnecessary clone() and childNodes() calls to the test. The current master had a segfault on the first round of GC. When this pull request had parsed 1,000 HTML documents, the RAM was steady at about 260Mb.

I think these tests will suffice to say that there aren't any memory leak issues added with this new memory management code.

defunctzombie:
looks good, please squash up the commit history

rc0x03:
Squashed into commit #4331075

rc0x03:
Thanks for merging. What's the ETA on publishing the next version on NPM?

src/libxmljs.cc
```diff
@@ -97,8 +97,26 @@ char* xmlMemoryStrdupWrap(const char* str)
...
+        // save a reference to the doc so we can still `unref` it
+        node->doc = xml_obj->doc;
+    }
+    return;
+}
+
 LibXMLJS::LibXMLJS()
 {
+    // set the callback for when a node is about to be freed
+    xmlDeregisterNodeDefault(xmlDeregisterNodeCallback);
```
triccardi-systran: What about [xmlThrDefDeregisterNodeDefault](http://www.xmlsoft.org/html/libxml-globals.html#xmlThrDefDeregisterNodeDefault) ? Are we guaranteed to always work in the same thread ?

PR #326, opened 2015-08-05
HTML toString() support

rc0x03:
This pull request causes Document.toString() to check whether the document is HTML or XML and serializes it accordingly. Also, Node.toString() now accepts a boolean argument for formatting the output. Since Document.toString has formatting turned on by default, Node.toString also formats the output by default.

Support for the rest of the xmlSaveOptions can be easily added. Maybe in a future release `toString` will also support an object argument like this:

```javascript
toString({
    format: true,       // formatting
    declaration: false,	// xml declaration
    emptyTags: true,    // include empty tags
    type: 'HTML'        // force output type
    xhtml: true,        // follow XHTML1 specific rules
})
```

### DIFF
--- a/src/xml_document.cc
+++ b/src/xml_document.cc
@@ -5,6 +5,8 @@
 
 #include <cstring>
 
+//#include <libxml/tree.h>
+#include <libxml/HTMLtree.h>
 #include <libxml/HTMLparser.h>
 #include <libxml/xmlschemas.h>
 #include <libxml/relaxng.h>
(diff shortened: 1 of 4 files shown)


rc0x03:
Note: the toString test will fail because it doesn't expect the string to be formatted by default

polotek:
Can you fix the tests?

rc0x03:
If this is added it should probably be in a major version due to Node.toString() formatting by default. Users may not expect formatted output and it could cause issues with some applications.

defunctzombie:
Can we make it not format by default to avoid major bump? Also, seems nice to have the other options or do you feel they are not needed? I would favor going the options object approach from the start even for a single boolean.

rc0x03:
Yes, I can make the Node.toString() not format by default. I just thought it might be worth considering for consistency between Node.toString() and Document.toString().

I can work on changing the code to accept an object, however I haven't worked with V8 objects in C++ before. It just needs to loop through the object, check each key to see if it matches an option name, then add the corresponding value to the `options` bitmask.

rc0x03:
Node.toString() now accepts an object with these keys:

###declaration
**XML_SAVE_NO_DECL**
True does nothing, false excludes the XML declaration. This doesn't seem to do anything in the case of Node.toString() because a declaration is only added for Document.toString(), but it's there just in case.

###format
**XML_SAVE_FORMAT**
True formats the output, false is the default.

###selfCloseEmpty
**XML_SAVE_NO_EMPTY**
True causes empty tags to be self-closing. `<node></node>` becomes `<node/>`. This option is only used in XML serialization. True by default.

###whitespace
**XML_SAVE_WSNONSIG**
True causes libxml to "format with non-significant whitespace". False by default.

###type
**XML** - Force HTML document to output XML format.
**HTML** - Force XML document to output HTML format.
**XHTML** - Force HTML output using "XHTML specific rules"

defunctzombie:
HOT

defunctzombie:
@rc0x03 can you please add a few tests for this additions and I will merge.

Nice stuff btw.. really good feature to have.

defunctzombie:
still needs tests and a rebase

rc0x03:
Working on tests

rc0x03:
Added tests and rebased

teleological:
These changes were merged in #341, so this PR can be closed.

PR #346, opened 2015-11-04
Fix memory leaks and segfaults

teleological:
Upgraded to v0.15 from v0.13, but then had to revert due to leaking memory. I think the root cause may be in #323 (43310754): Node wrappers ref() their parents, but don't unref() them in the case that they are unlinked from the parents. This could immortalize the parents, which in turn hold refs on the documents.

The fix (not ref'ing or unref'ing parents) is in e5996f4b, along with an alternate solution to the underlying problem: Not freeing a detached tree while it is still reachable from javascript. #323 accomplished this by preventing the root wrapper from being destroyed, but it only worked in the case that the detached root and an immediate child were wrapped, and in that precise order, such as in the "double_free" test in the ref_integrity test suite. The proposed solution allows an unreachable wrapper to be destroyed, but it only frees the wrapped node tree if no connected nodes are wrapped/reachable.

The cost of this solution — searching for reachable nodes whenever a wrapper is destroyed — is worth reconsidering. I made a couple optimizations:

1. If any of a wrapper's ancestors nodes are wrapped, the search is aborted. Such a node is reachable, so the tree cannot be freed.
2. If a wrapper's node is the document root, or is connected to the document root, the search is aborted, since the tree is reachable as long as the document is.

Further optimizations may be possible. Another approach would be to for the nodes of detached trees to maintain references to their root nodes.

Also included in this PR is ~~a fix for a memory leak when performing RelaxNG validation, explicit handles scopes (though I haven't seen any clear indication that leaking handles are a problem), and~~ exposing xmlMemUsed so that it can be used in tests to verify memory management.

v0.14.3 - v0.15 fail the "inaccessible_document_freed_after_middle_nodes_proxied" test (verifying the memory leak which prompted this PR) and all tests in the "referential_integrity" suite after "unlinked_tree_persistence_parent_proxied_first".

### DIFF
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -148,12 +148,8 @@ void xmlDeregisterNodeCallback(xmlNode* xml_obj)
     if (xml_obj->_private)
     {
         XmlNode* node = static_cast<XmlNode*>(xml_obj->_private);
-
-        // flag the XmlNode object as freed
-        node->freed = true;
-
-        // save a reference to the doc so we can still `unref` it
-        node->doc = xml_obj->doc;
+        node->xml_obj = NULL;
+        xml_obj->_private = NULL;
     }
     return;
 }
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -222,6 +222,12 @@ v8::Local<v8::Object> listFeatures() {
     return scope.Escape(target);
 }
 
+NAN_METHOD(XmlMemUsed)
+{
+  Nan::HandleScope scope;
+  return info.GetReturnValue().Set(Nan::New<v8::Int32>(xmlMemUsed()));
+}
+
 NAN_MODULE_INIT(init)
 {
       Nan::HandleScope scope;
@@ -241,6 +247,8 @@ NAN_MODULE_INIT(init)
       Nan::Set(target, Nan::New<v8::String>("features").ToLocalChecked(), listFeatures());
 
       Nan::Set(target, Nan::New<v8::String>("libxml").ToLocalChecked(), target);
+
+      Nan::SetMethod(target, "xmlMemUsed", XmlMemUsed);
 }
 
 NODE_MODULE(xmljs, init)
--- a/index.js
+++ b/index.js
@@ -34,3 +34,5 @@ var sax_parser = require('./lib/sax_parser');
 module.exports.SaxParser = sax_parser.SaxParser;
 module.exports.SaxPushParser = sax_parser.SaxPushParser;
 
+module.exports.xmlMemUsed = bindings.xmlMemUsed;
+
--- a/src/xml_node.cc
+++ b/src/xml_node.cc
@@ -229,64 +229,243 @@ XmlNode::New(xmlNode* node)
 }
 
 XmlNode::XmlNode(xmlNode* node) : xml_obj(node) {
-    this->freed = false;
     xml_obj->_private = this;
 
     // this will prevent the document from being cleaned up
     // we keep the document if any of the nodes attached to it are still alive
-    XmlDocument* doc = static_cast<XmlDocument*>(xml_obj->doc->_private);
+    this->doc = xml_obj->doc;
+    XmlDocument* doc = static_cast<XmlDocument*>(this->doc->_private);
     doc->ref();
+}
 
-    // check that there's a parent node with a _private pointer
-    // also check that the parent node isn't the doc since we already Ref() the doc once
-    if (xml_obj->parent != NULL &&
-        xml_obj->parent->_private != NULL &&
-        (void*)xml_obj->doc != (void*)xml_obj->parent)
-    {
-        static_cast<XmlNode*>(xml_obj->parent->_private)->Ref();
+/*
+ * Return the (non-document) root, or a proxied ancestor: whichever is closest
+ */
+xmlNode*
+findProxiedAncestorOrRoot(xmlNode *xml_obj) {
+    while ((xml_obj->parent != NULL) &&
+           (static_cast<void*>(xml_obj->doc) != static_cast<void*>(xml_obj->parent))  &&
+           (xml_obj->parent->_private == NULL)) {
+        xml_obj = xml_obj->parent;
     }
+    return ((xml_obj->parent != NULL) &&
+            (static_cast<void*>(xml_obj->doc) != static_cast<void*>(xml_obj->parent))) ?
+        xml_obj->parent : xml_obj;
+}
+
+
+/*
+ * Search linked list for javascript proxy, ignoring given node.
+ */
+xmlAttr*
+findProxiedAttrInList(xmlAttr *xml_obj, void *skip_xml_obj) {
+    xmlAttr *proxied_attr = NULL;
+    while (xml_obj != NULL) {
+        if ((xml_obj != skip_xml_obj) && (xml_obj->_private != NULL)) {
+            proxied_attr = xml_obj;
+            xml_obj = NULL;
+        }
+        else {
+            xml_obj = xml_obj->next;
+        }
+    }
+    return proxied_attr;
+}
+
+xmlNs*
+findProxiedNsInList(xmlNs *xml_obj, void *skip_xml_obj) {
+    xmlNs *proxied_ns = NULL;
+    while (xml_obj != NULL) {
+        if ((xml_obj != skip_xml_obj) && (xml_obj->_private != NULL)) {
+            proxied_ns = xml_obj;
+            xml_obj = NULL;
+        }
+        else {
+            xml_obj = xml_obj->next;
+        }
+    }
+    return proxied_ns;
+}
+
+xmlNode* findProxiedNodeInChildren(xmlNode *xml_obj, void *skip_xml_obj);
+
+/*
+ * Search document for javascript proxy, ignoring given node.
+ * Based on xmlFreeDoc.
+ */
+xmlNode*
+findProxiedNodeInDocument(xmlDoc *xml_obj, void *skip_xml_obj) {
+    xmlNode *proxied_node = NULL;
+    if ((xml_obj->extSubset != NULL) &&
+        (xml_obj->extSubset->_private != NULL) &&
+        (static_cast<void*>(xml_obj->extSubset) != skip_xml_obj)) {
+        proxied_node = reinterpret_cast<xmlNode*>(xml_obj->extSubset);
+    }
+    if ((proxied_node == NULL) &&
+        (xml_obj->intSubset != NULL) &&
+        (xml_obj->intSubset->_private != NULL) &&
+        (static_cast<void*>(xml_obj->intSubset) != skip_xml_obj)) {
+        proxied_node = reinterpret_cast<xmlNode*>(xml_obj->intSubset);
+    }
+    if ((proxied_node == NULL) && (xml_obj->children != NULL)) {
+        proxied_node =
+            findProxiedNodeInChildren(xml_obj->children, skip_xml_obj);
+    }
+    if ((proxied_node == NULL) && (xml_obj->oldNs != NULL)) {
+        proxied_node =
+            reinterpret_cast<xmlNode*>(findProxiedNsInList(xml_obj->oldNs, skip_xml_obj));
+
+    }
+    return proxied_node;
+}
+
+/*
+ * Search children of node for javascript proxy, ignoring given node.
+ * Based on xmlFreeNodeList.
+ */
+xmlNode*
+findProxiedNodeInChildren(xmlNode *xml_obj, xmlNode *skip_xml_obj) {
+
+    xmlNode* proxied_node = NULL;
+
+    if (xml_obj->type == XML_NAMESPACE_DECL) {
+        return reinterpret_cast<xmlNode*>(
+            findProxiedNsInList(reinterpret_cast<xmlNs*>(xml_obj), skip_xml_obj)
+        );
+    }
+
+    if ((xml_obj->type == XML_DOCUMENT_NODE) ||
+#ifdef LIBXML_DOCB_ENABLED
+        (xml_obj->type == XML_DOCB_DOCUMENT_NODE) ||
+#endif
+        (xml_obj->type == XML_HTML_DOCUMENT_NODE)) {
+        return findProxiedNodeInDocument(reinterpret_cast<xmlDoc*>(xml_obj), skip_xml_obj);
+    }
+
+    xmlNode *next;
+    while (xml_obj != NULL) {
+        next = xml_obj->next;
+
+        if ((xml_obj != skip_xml_obj) && (xml_obj->_private != NULL)) {
+            proxied_node = xml_obj;
+        }
+        else {
+
+            if ((xml_obj->children != NULL) && (xml_obj->type != XML_ENTITY_REF_NODE)) {
+                proxied_node = findProxiedNodeInChildren(xml_obj->children, skip_xml_obj);
+            }
+
+            if ((proxied_node == NULL) &&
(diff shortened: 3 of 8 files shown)


src/libxmljs.cc
```diff
@@ -173,7 +170,13 @@ v8::Local<v8::Object> listFeatures() {
     FEAT(ZLIB);
     FEAT(ICU);
     FEAT(LZMA);
-    return target;
+    return scope.Escape(target);
+}
+
+NAN_METHOD(XmlMemUsed)
+{
+  Nan::HandleScope scope;
```
defunctzombie: (style) spaces

defunctzombie:
Could you make a PR with the non-controversial changes (everything except the fix ref-counting and new export of xmlMemUsed) and I can merge those in. It will make the review and discussion around the actual memory issue fix easier to follow.

teleological:
Extracted non-controversial changes to #348. Thanks for looking at this!

defunctzombie:
@rc0x03 thoughts? since you are the author of the commits referenced here

rc0x03:
I'm looking into the code right now. My immediate concern is that this pull request will add some complexity and overhead to the ref issue. Ideally the solution wouldn't involve testing every child node.

If I understand correctly, the issue is that a node with a ref'ed parent can be taken away from that parent, therefore it can't unref the parent. Perhaps instead we can unref the parent any time a wrapped node is removed? This would require no added complexity or overhead. There may even be a libxml callback for that specific case.

I haven't experienced memory leaks in v0.15, so we should also be sure that the issue is in fact caused by immortal parent objects.

rc0x03:
Would it not be possible to keep ref'ing and unref'ing the parent wrappers and just add `parent->unref()` to any function that causes a node to become unlinked to its parent? So basically just by adding ` parent->Unref()` somewhere within `NAN_METHOD(XmlNode::Remove)`

rc0x03:
I'm going to try my suggested changes and run them against the tests you supplied to see if it's a working solution.

teleological:
The leak is reproduced in a test called "inaccessible_document_freed_after_middle_nodes_proxied" in the memory_management suite. It's not triggered by every use case, but only when:
1. A parent and child node are wrapped, in that order.
2. The child is unlinked from the parent.
In the case that the parent remains linked to the document, the document leaks as well.

Do I understand correctly that the parent-child refcount was introduced to solve the refint issue when the parent is unlinked and the parent wrapper is destroyed?

The main problem with the parent-child refcount approach, as I see it, is that it breaks down when the unparented root is a non-immediate ancestor of accessible/wrapped nodes. The "unlinked_tree_leaf_persistence_with_proxied_ancestor" test demonstrates that problem. If unparented nodes free their descendants when they become inaccessible/unwrapped, then it's necessary to keep them alive as long as they have have accessible descendants. Another issue is that you can only ref/unref wrapped nodes. This is tested in "unlinked_tree_persistence_child_proxied_first".

As mentioned above, an alternate approach is to have wrapped nodes in unlinked trees maintain a reference to the detached root. This approach is discussed in the lxml project:

https://github.com/lxml/lxml/blob/master/doc/memorymanagement.txt

In some ways, this would resemble what is currently done with node wrappers referencing the document wrapper. But there's some additional complexity since nodes in such trees would have to track the detached root, and operations which attach or unlink nodes would have to account for changes in the root, propagating them across accessible descendants, etc.

I'd be interested in exploring this approach, but it seems like it could get fairly involved.

teleological:
Reading the comments in libxml-ruby, it sounds like they have implemented that approach, keeping alive both documents and "top level, free standing nodes" which are either directly accessible, or accessible via descendants. Perhaps it could be used as a model?

teleological:
I briefly looked at the libmxml-ruby implementation, and it seems like it may actually have the same limitations at v0.14.3, i.e. only parents are marked as reachable, and only if they are wrapped. Perhaps I'm missing something?

On the other hand, the implementation of lxml looks similar to this pull request: `__dealloc__` calls a method called `attemptDeallocation`, which probes for the root then searches for wrapped descendants, aborting whenever it finds a wrapper.

teleological:
I'm in favor of renaming `xmlMemUsed`. Total failure of imagination there!

rc0x03:
This could be solved by recursively ref'ing the parents. The only problem is that each parent would need a ref count variable regardless of whether its wrapped. We would need a way to store a ref count in a parent xml_obj. This would ensure that each parent knows exactly how many wrapped XmlNodes it has.  Also, it would ref every parent recursively so it wouldn't depend on an ObjectWrap being a direct parent in order to be ref'ed.

I think if we can somehow get the refcount to be stored in the xml_obj it would be worth it because it's faster to go up the parent chain than down through every child. Looping through every child is especially impacting if there are hundreds of children and none of them are ObjectWraps.

In order to fix the memory leak issue, we would simply unref() the parents of any node that is being unlinked.

rc0x03:
If we do end up fixing this issue by looping through each child, I think the most efficient way would be to loop through each child of any node called in xmlDeregisterNodeDefault. Any wrapped children would be removed from the parent before the parent is freed, which would preserve the child. This would at least make it so we're looking for wrapped nodes while still freeing unwrapped nodes. The problem with this approach is that the wrapped children will become orphans. This would be fine if the expectation is that you lose any parents or documents that aren't kept alive by retaining a variable. However, I think parents and documents should stay around and for as long as the children. So this technique would perform better, but have some limitations.

rc0x03:
Just to clarify the issues at this point:

1. Children wrapped & ref'ed before a parent node has been wrapped

The child node will have no parent to ref(). Even if went through each parent until it hit the document, it would not find a `parent->_private` to ref() because one hasn't been created yet. Once a parent is wrapped it's refcount will be 0. The parent refcount should be 1, but `parent->_private` wasn't available at the time.

The symptom is that parents will not have updated ref counts and will be available for garbage collection as soon as they are unlinked from the document. Once they are unlinked, all of their children will be freed as well. If they are never removed from the document, the document will keep them alive and the issue does not occur.

2. Removed nodes won't unref their parent

The problem is that a wrapped node won't unref its parent if it is removed and no longer has a parent during garbage collection.

The symptom is that document ref count will never go down to 0 and will cause a memory leak. If the node isn't removed from the document, the issue does not occur.

3. We can only ref() wrapped parents

The problem is that a wrapped node needs to be able to increase the parent ref count regardless of whether `parent->_private->Ref()` is available.

The symptom is that an unwrapped parent can be garbage collected even if it contains a wrapped child. This only causes issues in cases where the parent is garbage collected before the document (i.e. only when the parent is removed from the document).

4. Looping through children is computationally expensive and time consuming

The problem is looping through the child nodes looking for any wrapped nodes. There is only ever one parent node, but there can be any number of child nodes.

The symptom is that the main thread to blocks for a while during garbage collection.

teleological:
It seems like the first strategy – ref'ing every parent – converges on wrapping every node. We'd be trading memory for CPU. I'm not sure it's a good trade, but it's worth further discussion. Maybe if would help to quantify the worst case for each approach.

I don't think piggybacking on the deregistration callback is feasible. By the time deregistration is invoked, it's too late to do the right thing in the case that wrapped descendants are discovered.
 
Whatever implementation is chosen, I'm mostly interested in seeing that these properties are observed:

1. Accessible nodes – those that are held as javascript references, or connected to nodes or documents which are held – are not freed.
2. Inaccessible nodes – those that are neither held as references nor connected to accessible nodes or documents – are freed.

rc0x03:
My idea for ref'ing every parent is to somehow store the refcount in the `xml_obj`. That way there is no requirement that _any_ node be wrapped in order to ref() a parent node. Perhaps it could be stored in [psvi](http://www.xmlsoft.org/html/libxml-tree.html#xmlNode) or [extra](http://www.xmlsoft.org/html/libxml-tree.html#xmlNode). Maybe even modify the actual `xmlNode` struct to have a `refs` property.

From what I understand, nothing is freed until the user defined deregistration callback has returned. So it could work as long as the children are unlinked before returning.

I agree, the solution should free whenever possible and you should always be able to retrieve a node's parent even if there are no javascript references.

rc0x03:
@defunctzombie what are your feelings on modifying the libxml source code so that `struct xmlNode` has `int refs`?

defunctzombie:
I would prefer not to mess with the libxml source but if it solves all the problems and remains performant then maybe it is for the best since my preference to avoid touching the source is to make patching easier but I never do that anyway :)

In general, I am happy with the solution you two decide upon here. I personally would favor one that does not add too much CPU and/or memory cost especially in the case where a user is not editing the XML but just parsing and traversing. I know folks use this library to build xml documents, etc but many also use it as a reliable way to scrape pages and slowing down the basic use cases for the less common ones doesn't seem ideal to me.

rc0x03:
I'll look at the libxml source code and see how `xmlNode`'s `extra` property is used. If it's possible to use that instead we can avoid modifying the source. Worse case scenario is that we do have to modify the source and someone eventually overwrites libxml/tree.h. It will be easy to remember to reinsert `int refs` because the binding code won't compile anymore. There will be an error like `xmlNode has no property "refs"`

Luckily this issue only effects code that will delete nodes from documents. The parent ref`ing solution seems to be the most efficient while covering all known edge cases and keeping the code relatively simple.

teleological:
@rc0x03: I look forward to seeing the implementation using `extra`. It looks like currently only the least significant 3 bits are used. I also note that it's not present in `xmlAttr` so attributes and their children may need special handling. How do you intend to initialize the reference count?

rc0x03:
@teleological very good point. We can't use it if it's not available in `xmlAttr`. I'd hate to handle refcounting differently depending on the node type. Looks like `xmlNode` and `xmlAttr` both have `psvi`. Maybe we should consider using that instead?

Here a scenario showing how it will work:

#####First a child node is wrapped
* A node is wrapped because a variable references it in javascript
* Let's assume it's the first node to be wrapped
* In the XmlNode constructor we call `XmlNode::recursiveRef()` (or whatever we call it)
* First it checks if the parent is wrapped (`parent->_private != NULL`)
* ##### If the parent is _not_ wrapped then:
* `recursiveRef` loops through each parent
* `parent->refs++`, `parent->parent->refs++`, etc.
* ##### if the parent _is_ wrapped then:
* `parent->_private->Ref()` to increase `ObjectWrap::ref_`
* `parent->_private->recursiveRef()` to continue from the parent
* The parent node will take it from there
* It will continue the ref`ing process up to the document

#####Then a parent node is wrapped
* A node is wrapped and it is a parent of the previously wrapped child
* The `XmlNode` constructor will see that `xml_obj->refs` is greater than 0
* The constructor will then set `XmlNode->refs_` to `xml_obj->refs`
* ObjectWrap now knows the actual ref count even though it didn't exist til now
* At this point, if any of this node's children are wrapped the will be modifying the _real_ parent refcount (i.e. `ObjectWrap::refs_`)

One more issue: if we have recursive ref`ing should it handle document ref`ing at the same time or should we continue to ref the document separately? Can you think of any edge cases where we wouldn't want recursive ref'ing to ref the document for us?

teleological:
Is it necessary to perform wrapper ref-counting if this alternate method of ref-counting is implemented and the destructor is aware of it?

rc0x03:
So if the destructor just doesn't free the node if the ref count is greator than 0? That's a good point. I haven't even thought about it. It's true that the nodes would be freed when the document is freed anyways. Although, what if a node is removed from the document, its refcount is greater than 0, and the destructor is called? The destructor would see that `xml_obj->parent == NULL` (since it was removed from the document), and call `xmlFreeNode` even though the refcount isn't 0. The children that ref'ed the node will be freed as well which would cause a problem. Even if the destructor checked that `xml_obj->parent == NULL` _and_ `refcount == 0`, there would still be an issue because once the destructor has been called the node is never checked again. That means that the node will hang around because it's not attached to the document anymore and it's not wrapped anymore. Although any child destructors will be called and its children will be freed, nothing is there to call `xmlFreeNode` on the parent anymore.

rc0x03:
Actually, we could probably implement an alternative using the `recursiveUnref()` function. When it unref's a parent down to refcount==0 _and_ the parent node isn't wrapped, it would free the parent node. I think that might work.

teleological:
A few thoughts:
* The assumption that you can free a detached (parent == null) node when its parent wrapper is destroyed is only valid if it there are no wrapped nodes in its descendants. Since that can only be assumed if the descendants hold actual references on the detached root – something that we're not currently contemplating – the assumption no longer holds.
* When propagating a "virtual" unreference upwards, you'd only want to take action – freeing a node and its descendants – when you reach a detached root. Attached roots can't be freed, since all the descending nodes are available through the document.
* You asked above should the proposed refcount scheme also handle "document ref'ing". I don't think it's up to that job for this reason: We need to keep the XmlDocument from being garbage collected for as long as nodes associated with the document are accessible – whether or not they are attached to the root.

rc0x03:
>Since that can only be assumed if the descendants hold actual references on the detached root

The child descendants will hold a reference on the detached root because it recursively increases each parent refcount. If a node has 2 wrapped children, then its refcount will be 2 and the document refcount will be 2. If there's another wrapped node somewhere in the document then the document refcount will be 3. Unless that other node is a child of the node with 2 wrapped children, then the wrapped children's parent refcount will still be 2 even though the document refcount is 3. So each node's refcount (linked or unlinked) will be equal to the number of wrapped children.

I think that our only option is to free any detached nodes with refcount==0, otherwise there would be a memory leak because the document won't free them. Also, any detached nodes will need to recursively subtract their refcount from each parent refcount, otherwise the document will leak because it has refs that won't be unref'd. The `replace()` function will also have to decrease the refcount of the old parent and increase the refcount of the new parent.

teleological:
I think I didn't make myself clear, re: descendants holding "actual references". I was attempting to respond to this portion of the comments:

>  The destructor would see that xml_obj->parent == NULL (since it was removed from the document), and call xmlFreeNode even though the refcount isn't 0. … Even if the destructor checked that xml_obj->parent == NULL and refcount == 0, there would still be an issue because once the destructor has been called the node is never checked again

I'm saying that there's no sense in preserving the rule that the wrapper destructor calls `xmlFreeNode` whenever `parent == NULL`, since the virtual refcount that you have described doesn't keep wrappers from being destroyed. On the other hand, it's not necessarily true that "the node is never checked again". You could and should also "free any detached nodes with refcount==0" in the wrapper destructor, but if the refcount is greater than zero, ownership passes to the wrapped descendants. When wrapped children are unlinked or when their wrappers are destroyed, you propagate an unref upwards. If you reach a root which is not connected to the document and the count is zero, you free the node.

I suspect we may have reached (or passed!) the point of diminishing returns in terms of discussing this approach without implementing it. Is it something that you're working on?

rc0x03:
I've almost got a working implementation. I just need to track down a few cases where the refcounts are subtracted down to -1.

rc0x03:
Almost done with the parent ref implementation. I'm going to clean it up a bit and remove the debugging code then I'll create a pull request so we can review it.

defunctzombie:
@rc0x03 @teleological I have access @polotek to add you both as collaborators; this should be done now.

rc0x03:
@defunctzombie thanks!

teleological:
@defunctzombie Thanks! @rc0x03 I'm looking forward to seeing what you've been up to.

rc0x03:
I plan to submit the pull request this afternoon. I've also exposed a `libxmljs::nodeCount()` function that returns the number of nodes that haven't been freed. It uses `xmlRegisterNode` and `xmlDeregisterNode` so the count doesn't rely on ObjectWrap. I've added a test at the end to make sure that the final nodeCount is 0 when all the tests have completed. With this test, any future pull requests won't pass if they cause a memory leak.

teleological:
This PR is superceded by #365

PR #347, opened 2015-11-04
Comment nodes should not escape xml characters

davydog187:
We want to allow the creation of comments that you would expect to see for IE conditionals in HTML markup. For example:

```html
<!--[if IE 6]>
Special instructions for IE 6 here
<![endif]-->
```

Should not escape the inner text of the comment. I can't think of a reason where escaping comment text *should* be necessary.

### DIFF
--- a/src/xml_comment.cc
+++ b/src/xml_comment.cc
@@ -19,8 +19,7 @@ NAN_METHOD(XmlComment::New) {
 
   // if we were created for an existing xml node, then we don't need
   // to create a new node on the document
-  if (info.Length() == 0)
-  {
+  if (info.Length() == 0) {
       return info.GetReturnValue().Set(info.Holder());
   }
 
@@ -28,17 +27,13 @@ NAN_METHOD(XmlComment::New) {
   assert(document);
 
   v8::Local<v8::Value> contentOpt;
-  if(info[1]->IsString()) {
+  if (info[1]->IsString()) {
       contentOpt = info[1];
   }
   v8::String::Utf8Value contentRaw(contentOpt);
   const char* content = (contentRaw.length()) ? *contentRaw : NULL;
 
-  xmlChar* encoded = content ? xmlEncodeSpecialChars(document->xml_obj, (const xmlChar*)content) : NULL;
-  xmlNode* comm = xmlNewDocComment(document->xml_obj,
-                                   encoded);
-  if (encoded)
-      xmlFree(encoded);
+  xmlNode* comm = xmlNewDocComment(document->xml_obj, (xmlChar *) content);
 
   XmlComment* comment = new XmlComment(comm);
   comm->_private = comment;
@@ -66,9 +61,7 @@ NAN_METHOD(XmlComment::Text) {
 
 void
 XmlComment::set_content(const char* content) {
-  xmlChar *encoded = xmlEncodeSpecialChars(xml_obj->doc, (const xmlChar*)content);
-  xmlNodeSetContent(xml_obj, encoded);
-  xmlFree(encoded);
+  xmlNodeSetContent(xml_obj, (xmlChar*) content);
 }
 
 v8::Local<v8::Value>
@@ -84,7 +77,6 @@ XmlComment::get_content() {
 
   return scope.Escape(Nan::New<v8::String>("").ToLocalChecked());
 }
-
 
 v8::Local<v8::Object>
 XmlComment::New(xmlNode* node)
--- a/test/comment.js
+++ b/test/comment.js
@@ -16,3 +16,21 @@ module.exports.text = function(assert) {
     assert.done();
 };
 
+
+module.exports.textWithSpecialCharacters = function(assert) {
+    var doc = libxml.Document();
+    var comm = libxml.Comment(doc);
+    var theText = 'my comment <has> special ch&r&cters';
+    comm.text(theText);
+    assert.equal(theText, comm.text());
+    assert.done();
+};
+
+module.exports.toStringWithSpecialCharacters = function(assert) {
+    var doc = libxml.Document();
+    var comm = libxml.Comment(doc);
+    var theText = 'my comment <has> special ch&r&cters';
+    comm.text(theText);
+    assert.equal("<!--" + theText + "-->", comm.toString());
+    assert.done();
+};


test/comment.js
```diff
@@ -16,3 +16,21 @@ module.exports.text = function(assert) {
     assert.done();
 };
 
+
+module.exports.text = function(assert) {
```
davydog187: This needs a different export name

src/xml_comment.cc
```diff
@@ -28,17 +28,13 @@ NAN_METHOD(XmlComment::New) {
...
+  if (info[1]->IsString()) {
       contentOpt = info[1];
   }
   v8::String::Utf8Value contentRaw(contentOpt);
   const char* content = (contentRaw.length()) ? *contentRaw : NULL;
 
-  xmlChar* encoded = content ? xmlEncodeSpecialChars(document->xml_obj, (const xmlChar*)content) : NULL;
-  xmlNode* comm = xmlNewDocComment(document->xml_obj,
-                                   encoded);
-  if (encoded)
```
davydog187: From my understanding encoded in was pointing to newly allocated memory so content does not need to be explicitly freed

stsvilik:
Need this fix badly, could you guys review (repo owners)?

defunctzombie:
Without encoding you are now making invalid xml no? In the tests, by leaving `<has>` as is, then the generated xml cannot be parsed back the same way as I understand it.

davydog187:
From the [XML W3 standard](http://www.w3.org/TR/REC-xml/#sec-comments)

> Within a CDATA section, only the CDEnd string is recognized as markup, so that left angle brackets and ampersands may occur in their literal form; they need not (and cannot) be escaped using " &lt; " and " &amp; ". CDATA sections cannot nest.

defunctzombie:
@lanoxx git history shows that you worked on implementing the comment support here. Any reason you chose to escape the special chars instead of just inserting the text as is?

davydog187:
To me, it appeared that the code for comment was copy pasted from `src/xml_element.cc`. I could be wrong, but they looked identical. May have been an artifact of that.

defunctzombie:
Makes sense as I didn't closely scrutinize the comment code when initially created.

Does libxml properly handle inserting `--` into the comment?

> For compatibility, the string " -- " (double-hyphen) must not occur within comments.]

Maybe we should have a test for that?

davydog187:
What would be the correct behavior for " -- "? Should it throw an error? Or should it escape the `--`

defunctzombie:
Dunno, seeing what happens when you try to insert it would be the first step. Maybe google around for how it is handled in libxml proper or other libs.

davydog187:
Currently, my change just allows it. For example, this test currently passes but probably shouldn't

```js
module.exports.doubleHyphen = function(assert) {
    var doc = libxml.Document();
    var comm = libxml.Comment(doc);
    var theText = 'my comment -- <has> special ch&r&cters';
    comm.text(theText);
    assert.equal("<!--" + theText + "-->", comm.toString());
    assert.done();
};
```

davydog187:
Googling around, it seems that there isn't a comprehensive answer to correctly handle this situation. [libvir](https://www.redhat.com/archives/libvir-list/2014-February/msg01278.html) chose to take `--` and replace it with `\-\-`. 

Some [stack overflow answers](http://stackoverflow.com/a/20885152) suggest replacing with the escaped ASCII or Unicode

defunctzombie:
Hm. What happens if you toString the whole document and then try to read it in again using this lib or maybe another lib. I am inclined to say that if parsing the resulting xml is handled by a few libs then maybe nothing needs to happen but that might also lead to more issues being raised down the line when someone else has behavior that breaks.

davydog187:
I'm inclined to say that the correct thing to do here is to unicode escape the slashes. That produces valid XML and does not escape the body of the comment in a non-obvious way.

What do you think?

davydog187:
The HTML escape of `-` is `&#45;`

davydog187:
@defunctzombie would appreciate input so we can push this through

rc0x03:
+1 for escaping `-`

src/xml_comment.cc
rc0x03: [\[the\] result must be deallocated.](http://www.xmlsoft.org/html/libxml-entities.html#xmlEncodeSpecialChars)

rc0x03: `xmlEncodeSpecialChars` escapes '<', '>', '"' and '&'

rc0x03:
@polotek I'm torn between accuracy and compatibility. I'm leaning more towards compatibility though.

I think that leaving in `xmlEncodeSpecialChars` is okay because it only encodes <, >, &, and '\r'. Leaving in these characters could cause more problems than they solve.

We could always create an option that says not to encode comment content. Also, maybe `comment.text()` should decode the entities if it encoded them? That way the user gets back exactly what they put in.

rc0x03:
If we do escape comment context, then we should escape any `-` as well.

The comment may have text, with the additional restriction that the text must not start with a single ">" (U+003E) character, nor start with a U+002D HYPHEN-MINUS character (-) followed by a ">" (U+003E) character, nor contain two consecutive U+002D HYPHEN-MINUS characters (--), nor end with a U+002D HYPHEN-MINUS character (-).

(http://www.w3.org/TR/html5/syntax.html#comments)

defunctzombie:
I am a bit surprised that libxml itself allows you to create an invalid xml document in this way. I would have expected their default behavior (or flag behavior) to allow for setting text that correctly maintains the validity of the xml document.

stsvilik:
Just throwing in my 2c - if something is considered invalid according to spec, it must throw an error. That would be cheaper then trying to escape and also delegates responsibility of doing the "right" thing to a user :)

rc0x03:
@stsvilik that's true. After thinking about, getting rid of the `xmlEncodeSpecialChars` might be the best way to go (as long as libxml can parse whatever it puts out). We still need to escape `--` and we also can't have `--->`.

davydog187:
I will update to add an escaping function for comments.

davydog187:
As far as I can tell, libxml does not provide one.

stsvilik:
My point was that escaping (of ```--``` should be handled by a user and not this library. 

Throwing an error when user doesn't escape invalid characters would:
- educate the user
- let them handle the processing of the comment value so there are no surprises on the output
- save us lines of unnecessary code

defunctzombie:
I am ok with throwing or just having it allow invalid xml since libxml seems to do that. The goal of this library is to wrap libxml and not necessarily fix the warts of libxml.

rc0x03:
:+1: @defunctzombie I like that philosophy. It makes these decisions much easier. I don't think it should throw an error.

My final vote is get rid of any escaping. Maybe leave a comment saying what should be escaped according to standards and why we're not escaping it. That way we can refer to the comment if issues appear in the future. It will also remind future contributors that libxmljs shouldn't be a patch for libxml.

rc0x03:
Sorry we had to make this issue so roundabout :p. At least we explored all the issues. I'd say this is ready to be merged once you're done with the changes and squash the commits.

src/xml_comment.cc
davydog187: I am no longer using `xmlEncodeSpecialChars`, so there should be no reason to deallocate

davydog187:
Squashed my commits. I'm not sure if I was supposed to change anything else, the conversation became confusing :smile: 

Right now I am *not* escaping comment text in any way. If this is the intended implementation then this should be good to merge. @rc0x03 can I have your input?

davydog187:
Seeing that there are currently no objections, I think this is good to merge as is.

PR #350, opened 2015-11-06
Document.fromHtmlFragment

davydog187:
This PR helps alleviate the issue of wanting to parse HTML fragments without them being wrapped in implied document elements (doctype, head, body, etc)

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ module.exports.parseXml = Document.fromXml;
 
 /// parse an html string and return a Document
 module.exports.parseHtml = Document.fromHtml;
+module.exports.parseHtmlFragment = Document.fromHtmlFragment;
 
 // constants
 module.exports.version = require('./package.json').version;
@@ -26,7 +27,7 @@ module.exports.Comment = require('./lib/comment');
 
 // Compatibility synonyms
 Document.fromXmlString = Document.fromXml;
-Document.fromHtmlString = Document.fromHtmlString;
+Document.fromHtmlString = Document.fromHtml;
 module.exports.parseXmlString = module.exports.parseXml;
 module.exports.parseHtmlString = module.exports.parseHtml;
 
--- a/lib/document.js
+++ b/lib/document.js
@@ -143,6 +143,23 @@ module.exports.fromHtml = function(string, opts) {
     return bindings.fromHtml(string, opts);
 };
 
+/// parse a string into a html document fragment
+/// @param string html string to parse
+/// @param {encoding:string, baseUrl:string} opts html string to parse
+/// @return a Document
+module.exports.fromHtmlFragment = function(string, opts) {
+    opts = opts || {};
+
+    // if for some reason user did not specify an object for the options
+    if (typeof(opts) !== 'object') {
+        throw new Error('fromHtmlFragment options must be an object');
+    }
+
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -26,7 +26,6 @@ module.exports.Comment = require('./lib/comment');
 
 // Compatibility synonyms
 Document.fromXmlString = Document.fromXml;
-Document.fromHtmlString = Document.fromHtmlString;
```
davydog187: assigning nothing to nothing won't help anyone

src/xml_document.cc
```diff
@@ -244,17 +246,25 @@ NAN_METHOD(XmlDocument::FromHtml)
     xmlResetLastError();
     xmlSetStructuredErrorFunc(reinterpret_cast<void*>(&errors), XmlSyntaxError::PushToArray);
 
+
+    static const int DEFAULT_PARSING_OPTS = 0;
```
davydog187: wasn't sure of the best practice for placing constants. I can put these in the header file if preferred

lib/document.js
```diff
@@ -139,6 +139,22 @@ module.exports.fromHtml = function(string, opts) {
...
+/// parse a string into a html document fragment
+/// @param string html string to parse
+/// @param {encoding:string, baseUrl:string} opts html string to parse
+/// @return a Document
+module.exports.fromHtmlFragment = function(string, opts) {
+    opts = opts || {};
+    opts.excludeImpliedElements = true;
+
+    // if for some reason user did not specify an object for the options
+    if (typeof(opts) !== 'object') {
```
defunctzombie: this check should be before the use of `opts`

index.js
defunctzombie: I think this was meant to be `= Document.parseHtml`

teleological: My vote would be for `fromHtml`, to keep the parallel:
~~~~
Document.fromHtmlString = Document.fromHtml;
~~~~

src/xml_document.cc
teleological: Seems to me like they're unlikely to be reused, but I think they'd fit better as static members of the class. If they are moved, header will need to include "libxml/HTMLparser.h"

teleological:
@defunctzombie Added my comments. Looks good, but would be nice to have `fromHtmlString` replaced with something working.

index.js
defunctzombie: Yea, sorry, this is what I meant. I think the current line was a typo and was meant to be what you have written.

rc0x03:
@defunctzombie This looks good to me after the suggested changes have been applied. It would be really nice to have all of the [htmlParserOptions](http://www.xmlsoft.org/html/libxml-HTMLparser.html#htmlParserOption) available at some point.

index.js
davydog187: Changed to @teleological suggestion

davydog187:
@rc0x03 @teleological @defunctzombie I think this should be good to go based on your feedback

teleological:
:+1: Looks ready for squash and merge to me.

davydog187:
Squashed

rc0x03:
:+1:

davydog187:
@defunctzombie can this be merged? We would love to start using this feature in our apps :)

PR #265, opened 2014-10-20
Better support for text nodes

polotek:
This is branch to add better support for adding text nodes per #263. The task list below outlines the steps needed for completion.

- [x] add `addText` method to append text nodes to an element.
- [x] add `addPrevText` and `addNextText`

Issue #263, opened 2014-10-11
Missing ability to create plain text nodes


lillem4n:
Hello!

First, thanks for a great lib! I just replaced xmldom with this for a major performance boost! :D

However, I'm missing a very important part for my project; creating text nodes. As far as I understand I cannot change this: <foo>bar <a>some</a></foo> to <foo>bar <a>some</a> lol</foo>, since the new node is a plain text node. I believe the binding I need is this: http://xmlsoft.org/html/libxml-tree.html#xmlNewDocText

If I am mistaken, gain some stackoverflow points by answering my question here: http://stackoverflow.com/questions/26289359/how-to-create-text-node-in-libxmljs

Thanks alot! :)

hildjj:
Any status on this?  I need something like addNextSibling in order to add text after a given node.  Use case: `<foo xmlns:xi='http://www.w3.org/2001/XInclude'>boo <xi:include href='testing' parse='text'/> baz</foo>`.  Alternately, in this case a binding to `xmlReplaceNode` that accepts text, element, or CDATA nodes would work for me.

RobinQu:
+1

polotek:
It looks like we never landed the attached PR. I'm going to fix it up and get it landed later this week.

RobinQu:
Bravo!

teleological:
Implemented in #351

PR #351, opened 2015-11-06
Expose text node class

davydog187:
Taking a look at #265, it seemed that there could be better support for Text nodes by exposing a `Text` class. This PR does just that.

I will continue to improve the test coverage and review the features exposed. This PR is to get the discussion started

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -15,6 +15,7 @@
         'src/xml_node.cc',
         'src/xml_sax_parser.cc',
         'src/xml_syntax_error.cc',
+        'src/xml_text.cc',
         'src/xml_xpath_context.cc',
       ],
       'dependencies': [
--- a/index.js
+++ b/index.js
@@ -20,9 +20,10 @@ module.exports.libxml_debug_enabled = bindings.libxml_debug_enabled;
 module.exports.features = bindings.features;
 
 // lib exports
+module.exports.Comment = require('./lib/comment');
 module.exports.Document = Document;
 module.exports.Element = require('./lib/element');
-module.exports.Comment = require('./lib/comment');
+module.exports.Text = require('./lib/text');
 
 // Compatibility synonyms
 Document.fromXmlString = Document.fromXml;
--- a/lib/comment.js
+++ b/lib/comment.js
@@ -14,8 +14,7 @@ var Comment = function(doc, content) {
                         'instance of Document');
     }
 
-    var comm = new bindings.Comment(doc, content);
-    return comm;
+    return new bindings.Comment(doc, content);
 };
 
 Comment.prototype = bindings.Comment.prototype;
--- a/lib/document.js
+++ b/lib/document.js
@@ -12,12 +12,12 @@ function assertRoot(doc) {
 /// @param {string} version xml version, default 1.0
 /// @param {string} encoding the encoding, default utf8
 /// @constructor
-var Document = function(version, encoding) {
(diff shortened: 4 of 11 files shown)


stsvilik:
So implementation signature suggests we can do something like this?

```js
var newTextNode = new libxml.Text(document, "my text");
```
Where we can later append new text node to another node

```js
var newTextNode = new libxml.Text(document, "my text");
var newElement = new libxml.Element("div");
newElement.addChild(newTextNode);
```
and in relation to #265 we could also do
```js
var newTextNode = new libxml.Text(document, "my text");
var newElement = new libxml.Element("div");

//To let text follow the tag ex.: <div></div>my text
newElement.addPrevSibling(newTextNode);
//or to let text precede tag ex.: my text<div></div>
newElement.addNextSibling(newTextNode);
```
Is this correct?

davydog187:
Added a test to prove your first test case works.

davydog187:
What does @polotek think?

davydog187:
Also validated your second question.

davydog187:
This should also address #263

rc0x03:
Why not inherit from `XmlElement` instead of implementing most of its functions separately? This seems like it could cause consistency issues in the future. It would also significantly reduce the size of the code.

rc0x03:
Also, if it does inherit from `XmlElement` then we need to decide what inapplicable things like `node->name()` should do. They could be removed and unimplimented. They could throw an error if it's a text node. They could do nothing or return an empty string. They could do what their DOM equivalents would do, for example `text_node.nodeName` in a browser would be `"#text"`. Maybe @polotek or  @teleological can offer suggestions for that.

rc0x03:
Actually, since a libxml text node is the same thing as a regular node then `XmlElement` might be able to handle inapplicable functions like `node->attrs()` automatically without causing issues.

stsvilik:
IMHO the difference between XmlNode and XmlElement is significant and most of the exposed methods do not apply to Text, for starters you can not append anything to a text, Text has no attributes nor does it need a name method. Most of the extra weight which comes with XmlElement is extraneous.

rc0x03:
@stsvilik great points. I just wish the shared functions were all in one location. Especially since garbage collection reference counting might be added to functions like `remove` and `replace` in the future.

Also, you may need to add a `node->type == "text"` case inside `XmlNode::New`

davydog187:
@rc0x03 I'd be happy to work towards removing duplicate logic. If this was javascript I'd share them by way of a mixin. In C++, there are also facilities to provide mixins but I'm not sure how well they would play with `NAN`. Any suggestions?

davydog187:
@rc0x03 also the node type is not necessary. If you take a look at my tests you'll see that its already been validated.

Possibly one way to share some of this functionality is to move some of the duplicated code up into XmlNode? I haven't validated this would work yet, but it could be an option

stsvilik:
I kinda need this feature - is anything besides a hypothetical refactoring of shared traits preventing this feature from going in? I feel that refactoring could wait while this feature might as well be used now.

rc0x03:
@stsvilik I agree, this can be merged as-is for now

stsvilik:
:+1:  lets do it :)

stsvilik:
Please create new release with these changes

defunctzombie:
@stsvilik I am waiting on two more PRs to finalize then a release. If you really need a feature you can depend directly on a github repo from npm.

Issue #232, opened 2014-04-23
nodejs crashed with SIGSEGV in libxmljs::XmlNamespace::~XmlNamespace()


lanoxx:
Here is a similar almost identical version of the show case from the constructor bug. This version segfaults in the destructor instead of the constructor:

    var xml = require('libxmljs');

    var NAMESPACE1 = "http://niceNamespace.org";
    var PREFIX1 = "someThingNice";
    var NAMESPACE2 = "http://interesting.org";
    var PREFIX2 = "interesing";
    var NAMESPACE3 = "http://funnyname.org";
    var PREFIX3 = "funny";

    // Create 20 documents
    for (var i=0; i<20; i++) {
        var doc = xml.Document();
        var rootElement = xml.Element(doc, 'rootElement' + i, '');
        rootElement.defineNamespace(PREFIX1, NAMESPACE1);
        rootElement.defineNamespace(NAMESPACE2);
        rootElement.defineNamespace(PREFIX3, NAMESPACE3);
        rootElement.namespace(PREFIX1, NAMESPACE1);
        doc.root(rootElement);
        // Put 100 child elements into the rootElement
        for (var j = 0; j < 100; j++) {
            var child1Element = xml.Element(doc, 'child1', '');
            child1Element.namespace(PREFIX2, NAMESPACE2);
            rootElement.addChild(child1Element);
            //Put 5 child elements into the second level of each child element
            for(var k = 0; k < 5; k++) {
                var child2Element = xml.Element(doc, 'child2', '');
                child2Element.namespace(PREFIX3, NAMESPACE3);
                child2Element.text("foobar");
                child2Element.attr('tom');
                child1Element.addChild(child2Element);
                var comment = xml.Comment(doc, 'Some short comment.');
                child1Element.addChild(comment);
            }
            child1Element.child(3).remove(); //Remove the third second level child.
        }
        rootElement.child(0).remove(); //ERROR
    }

teleological:
This is different than #163, but it is fixed by #352.

defunctzombie:
closing, future discussions in the PR.

PR #352, opened 2015-11-07
Prevent segfault when freeing namespace wrapper

teleological:
~XmlNamespace segfaults if the underlying xmlNs has been freed. Enhanced deregister node callback to prevent this from happening.

### DIFF
--- a/src/libxmljs.cc
+++ b/src/libxmljs.cc
@@ -8,6 +8,7 @@
 #include "xml_document.h"
 #include "xml_node.h"
 #include "xml_sax_parser.h"
+#include "xml_namespace.h"
 
 namespace libxmljs {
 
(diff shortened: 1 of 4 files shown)


defunctzombie:
@rc0x03 ping since you have been more involved with the code lately I'd appreciate a review

defunctzombie:
@rc0x03 @teleological I think both of you are having great contributions and discussions around the issues in this repo. I am happy to go with the recommended approach you two come to as long as PRs have tests. So just let me know when you both are happy with the PR and I can merge it. After a few of these I am also happy to add you as collaborators so that I don't need to have the final say.

defunctzombie:
@teleological can you explain a bit (ideally via a comment in the code) why this approach must be taken over maybe a simpler approach (the one we currently had that seems to fail). Would be nice to annotate why looping and the various checks are required at a high level.

teleological:
@defunctzombie Just to be clear, did you want to see more comments in this PR (#356) or the other one (#346)? Thanks again for your work in keeping this project moving.

teleological:
Added a comment explaining why deregistration is needed for nodes and namespaces.

teleological:
Also added comments explaining why context is conditionally unref'd.

teleological:
This PR fixes #232. The script provided in that bug report no longer segfaults.

rc0x03:
@defunctzombie this looks good to me. The only thing I'm wondering is if `xmlNs->context` is always a Document. If it can sometimes be a Node then we should check for that and `static_cast<XmlNode*>`.

teleological:
The assumption that `context->_private` is an `XmlDocument` predates this PR. (However, it seems like  defensible one: `context` is typed as an `xmlDoc` pointer.)

src/libxmljs.cc
```diff
@@ -97,9 +98,53 @@ char* xmlMemoryStrdupWrap(const char* str)
...
+            wrapper->xml_obj = NULL;
+            ns->_private = NULL;
+        }
+        ns = ns->next;
+    }
+}
+
+void deregisterNodeNamespaces(xmlNode* xml_obj)
+{
+    xmlNs *ns = NULL;
```
defunctzombie: Inconsistent pointer placement. Other locations have the pointer next to the var name.

Maybe adding: http://clang.llvm.org/docs/ClangFormat.html to the project would be nice?

defunctzombie:
Looks good minus one nitpick on style inconsistency. Ping me once updated and I will merge.

src/libxmljs.cc
teleological: @defunctzombie – Funny! I actually changed my commit, after the initial push, from using `type *var` to `type* var`, because I saw the latter form is used exclusively in this file, and because I assumed my preference for the former style is a hangover from plain old C. I missed one usage: That first variable declared in `deregisterNodeNamespaces`.

I pushed a fix so that line is consistent, and now all pointer declarations in the file use asterisk next to type, rather than next to the variable name. I'm content to follow other convention, but I note that if I rewrite the commit to use `type *var`, they will no longer match the declarations in the rest of the file.

I didn't know about `clang-format`. Looks like a good thing.

Issue #272, opened 2014-11-05
XmlDocument should be subclass of XmlNode


hildjj:
This is quite typical in DOMs.  The most important method to get access to is `type()`

Alternately, just add `type`, please, so I can process an entire document with the same recursive code.

teleological:
Added `type()` in #353.

defunctzombie:
Referenced PR merged. Docs need updating.

teleological:
@defunctzombie Docs updated. I think this can be closed out.

PR #353, opened 2015-11-07
Add type() to document per #272

teleological:


defunctzombie:
needs wiki docs update plz

PR #354, opened 2015-11-08
Don't leak namespace array

teleological:
~~This can be tested once/if xmlMemUsed is exposed per #346.~~

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,3 +34,5 @@ var sax_parser = require('./lib/sax_parser');
 module.exports.SaxParser = sax_parser.SaxParser;
 module.exports.SaxPushParser = sax_parser.SaxPushParser;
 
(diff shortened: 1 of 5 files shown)


rc0x03:
Can will just merge in xmlMemUsed via a separate pull request? I think we all agree it's a good feature. I think we should take after `process.memoryUsage()` and just make it `libxml.memoryUsage()` for consistency.

teleological:
Cherry-picked the `xmlMemUsed` commit and renamed it to `memoryUsage`. I found another problem with the way namespace memory is handled and will be pushing another commit shortly.

teleological:
It turns out that the other issues are pretty broad and don't directly impact this PR. So it could, barring objections, be merged.

Issue #226, opened 2014-04-15
segmentation fault with libxmljs


SaltwaterC:
I was checking for [XXE](https://www.owasp.org/index.php/XML_External_Entity_%28XXE%29_Processing), even though it shouldn't be possible with libxml2 2.9 used by OS X 10.9.2. Used node.js v0.10.26 and libxmljs v0.9.0.

The code used to trigger the segfault:

``` javascript
// xxe.js
var parser = require('libxml-to-js');
var xml = require('fs').readFileSync('./xxe.xml').toString()
parser(xml, function (err, res) {
    if (err) {
        console.error(err)
        process.exit(1)
    }
    console.log(res)
})
```

``` xml
<!-- xxe.xml -->
<?xml version="1.0" encoding="ISO-8859-1"?>
 <!DOCTYPE foo [
   <!ELEMENT foo ANY >
   <!ENTITY xxe SYSTEM "file:///etc/passwd" >]><foo>&xxe;</foo>
```

``` bash
node xxe.js
[1]    28749 segmentation fault  node xxe.js
```

[libxml-to-js](https://github.com/SaltwaterC/libxml-to-js) is just a wrapper for libxmljs that passes back a JavaScript object following the XML doc structure.

teleological:
Fixed in #359

rc0x03:
@teleological I think we can close this if you think it's been fixed by now. It's been more than a year and the memory management situation has evolved quite a bit in that time.

teleological:
@rc0x03 The problem here is not due to memory management, but due to treating non-element nodes as elements,  and applying inapplicable operations to the wrapped nodes. I will clean up #359, which will address this issue.

samskeller:
Is there any chance this can get fixed? I use `libxmljs` via the `libxml-to-js` package and the segmentation faults from this issue are causing us some problems. Thanks!

PR #359, opened 2015-11-15
Refactor element/text base class

teleological:
Factored out common methods in Element/Text. ~~Couldn't come up with a decent name for the result; would welcome suggestions.~~

~~Perhaps more importantly: Noticed some memory issues when adding text siblings. It may be good to patch those before letting this out into the wild. Working on a fix, but wanted to get a clean refactor out first.~~ ( Bug fixes in #362 )

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -9,6 +9,7 @@
         'src/libxmljs.cc',
         'src/xml_attribute.cc',
         'src/xml_document.cc',
+        'src/xml_fraternal_node.cc',
         'src/xml_element.cc',
         'src/xml_comment.cc',
         'src/xml_namespace.cc',
--- a/binding.gyp
+++ b/binding.gyp
@@ -9,7 +9,7 @@
         'src/libxmljs.cc',
         'src/xml_attribute.cc',
         'src/xml_document.cc',
-        'src/xml_fraternal_node.cc',
+        'src/xml_non_attribute_node.cc',
         'src/xml_element.cc',
         'src/xml_comment.cc',
         'src/xml_namespace.cc',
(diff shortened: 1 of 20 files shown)


teleological:
Added a fix for when text nodes are added as children to nodes whose final children are also text nodes. Fixes are needed for adding next and previous siblings as well.

teleological:
Added fix and tests for adding merged text sibilings. All merged text tests segfault without this PR.

teleological:
Renamed "fraternal" nodes to "non-attribute nodes", updated XmlComment to be a subclass, and changed XmlNode::New so that it creates the correct kind of subclasses (fixes #226)

rc0x03:
It could also be called a proto-element type. Also, in #360 the `add_child`, `add_next_sibling`, `add_prev_sibling`, and `import_element` functions have all been moved to `XmlNode`. The reason I moved these to XmlNode is so that the refcount logic is all in one location. This way, anything that inherits from XmlNode will have these functions available, but won't have them exposed via the API unless the XmlNode inheritor implements `addChild()`, etc.

teleological:
@rc0x03 I also moved `import_element` to `XmlNode`, but didn't move the others because they didn't seem a good fit with `XmlAttribute`, which also descends from node. Have you looked into that? I've started to review #360.

teleological:
I'm going to extract the memory fix for adding/merging text nodes into a separate PR that can be merged without delay. I will leave the refactoring of the Element/Text base class for later consideration. The correction of the wrapper selection for child nodes, etc. is technically a bug fix, but it also impacts usage, since nodes returned from some operations won't respond to the same API as before the change. So it would be kinder not to shoehorn it into a minor release.

teleological:
Bug fixes extracted to #359

PR #360, opened 2015-11-15
Memory management: recursively reference each parent xmlNode

rc0x03:
Changes made by this pull request:
 * `struct xmlNode` now has a `refs` property
 * XmlNodes no longer use `ObjectWrap::Ref` or  `ObjectWrap::Unref`
 * Reference counts are stored in `struct xmlNode`
 * XmlElement functions that require refcount logic have been moved to XmlNode
 * Various XML modification functions such as `remove` and `add_child` now have refcount logic
 * `libxmljs.nodeCount()` returns the number of nodes still in memory
 * Added test to make sure there are no remaining nodes after all tests have completed

### DIFF
--- a/src/xml_node.cc
+++ b/src/xml_node.cc
@@ -81,7 +81,7 @@ NAN_METHOD(XmlNode::Namespace) {
   }
 
   node->set_namespace(ns->xml_obj);
-  return info.GetReturnValue().Set(info.Holder());
+  return info.GetReturnValue().Set(node->get_namespace());
 }
 
 NAN_METHOD(XmlNode::Namespaces) {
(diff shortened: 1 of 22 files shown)


teleological:
I will continue to review tomorrow. One direction that needs further study is interplay with libxml functions that mutate the graph.

rc0x03:
@teleological Thanks for the in-depth review

teleological:
It looks like however we resolve some issues – especially those related to `addChild` – we will break backwards compatibility, so a major version bump is probably in the works. But before we move to a new version, I think we should, as far as possible, fix the issue on 0.* without forcing an upgrade to 1.*. Which is to say, I think we separate backwards-compatible changes from those which aren't.

So for example, `import_element` should not eagerly copy in 0.* – but I don't think it really needs to do so, since the effect of not copying with the new refcount system is not worse than the status quo: The risk of leaving zombie nodes in the tree when freeing wrappers is the same. Or am I missing something?

src/xml_node.cc
```diff
@@ -230,68 +230,137 @@ XmlNode::New(xmlNode* node)
...
+    while (node->parent != NULL) {
+        node = node->parent;
+        if (node->type == XML_DOCUMENT_NODE || node->type == XML_HTML_DOCUMENT_NODE)
+            break;
+        node->refs += count;
+    }
+}
+
+// recursively unref each parent
+void XmlNode::unref(int count) {
```
davydog187: I'm a little worried about this approach. Rather than explicitly counting references, why not hold an object which tracks referencing, similar to a smart pointer. This would be much safer than explicitly counting

PR #361, opened 2015-11-16
Improve Windows support

gagern:
libxmljs won't build with Visual Studio 2015 because that will complain that `snprintf` was defined as a macro. Other code, [node-libxslt](https://github.com/albanm/node-libxslt) in this case, can't link against libxmljs on Windows since the underlying libxml api isn't exported by the node extension due to the static library step in between.

Both these issues are discussed in https://github.com/albanm/node-libxslt/issues/27 and adressed in my fork [libxmljs-mt](https://www.npmjs.com/package/libxmljs-mt). Here I've cherry-picked those portions which don't relate to #296 or #297. In particular, the commits in this request here are cherry-picked from https://github.com/gagern/libxmljs/commit/6967c09cd2f55c7d0d91c19595c4a07158d43acc and https://github.com/gagern/libxmljs/commit/12213fff63c38b2cf38cf9fa42e87149399ad2c4. I've taken a slightly more backwards-compatible approach to the `snprintf` macro here, though.

Note that I've also set up [AppVeyor](http://www.appveyor.com/) for this project. You may choose to do the same if you want to have CI on Windows as well. If not, the `appveyor.yml` is just another file in your repo, useful if any developer needs to try things for Windows without being on Windows themselves.


package.json
```diff
@@ -2,12 +2,13 @@
   "name": "libxmljs",
   "author": "Marco Rogers",
   "contributors": [
-    "Jeff Smick"
+    "Jeff Smick",
+    "Martin von Gagern <Martin.vGagern@gmx.net>"
```
defunctzombie: please don't add yourself without consulting us ;)

package.json
```diff
@@ -2,12 +2,13 @@
...
   ],
   "description": "libxml bindings for v8 javascript engine",
   "version": "0.15.0",
   "scripts": {
-    "test": "node --expose_gc ./node_modules/.bin/nodeunit test"
+    "test": "node --expose_gc node_modules/nodeunit/bin/nodeunit test"
```
defunctzombie: why was this changed? Is this not available on windows? Actually a better thing to do is just to have `nodeunit test` here since `npm test` will automatically setup the correct path.

vendor/libxml/libxml.gyp
```diff
@@ -1,56 +0,0 @@
-{
```
defunctzombie: why does this file need to be removed? This built on windows just fine previously. Can we just alter something in this file instead of copying it to the root gyp file?

defunctzombie:
Seems that there are multiple issues at work in this PR. One is the snprintf macro and others are related to some build issues. Ideally the PR focus is kept at the specific issues at hand and free of extra noise.

Is AppVeyor free for OSS?

package.json
defunctzombie: Cause this isn't where this project tracks contributors. I could just remove this entry altogether. We don't really have a separate contributors file etc for tracking. A git log on the repo or a git blame can show who is responsible for which file.

Maybe people make contributions and they are all appreciated. Attribution is provided by your commits :)

gagern:
Since there was no reply as to whether you really want this split and the context lost, I now opened #386 and #387 to discuss two of the core issues here. When you get AppVeyor up and running, you probably will notice two remaining issues.

One is the fact that your [libxml upgrade](https://github.com/polotek/libxmljs/commit/9acfcaf7d736ec64656bce04259e3b526ecc4b5d#diff-249d2ce676689422cc0e3f3b3c87deac) removed the `#define snprintf _snprintf` which is required by some versions of Visual Studio but forbidden by others. Since #384 added the other removed preprocessor macros in a different file (which in my opinion is a good thing), I'll leave it to you to decide where to put the `snprintf` handling.

Another issue is that at least in my experience, running the `npm test` command will require a different path to the `nodeunit` script. As you were unhappy with my solution to that, I'll leave it to you to come up with something you like better.

Feel free to get inspiration for both these issues from the cmmits and diffs here. But I don't feel like porting any more of my changes from my fork just now, so please don't ask for pull requests for three lines of code.

package.json
Klortho: All due respect, but if somebody makes substantial contributions, then they should be listed in the package.json as a contributor. It seems just a little bit petty to insist his name should be kept off.

defunctzombie: Im glad you have a strong opinion on this. Like I said, this field **is not actively used by us** and therefore I don't feel that updating it is required. Additionally, do you see my name? Or the name of any other contributor? And furthermore. The changes have not been merged in or contributions added. I do not appreciate when folks automatically add themselves to such lists (when they are maintained). These lists are for project collaborators to decide who is added. The new maintainers are free to revive this list and determine rules for keeping it up to date but if you need your name listed in a file to allow your contributions to be accepted, then maybe you are contributing to a project for the wrong reasons.

Klortho: Sorry, maybe you misunderstood me in a couple of places. Let me try to clarify.

- I don't have any strong opinions, I was just sharing my two cents.
- You said you "don't feel that updating it is required". But, it's not required to *not update* it, right?
- I didn't notice your name wasn't on it. If that's the sticking point, why don't you add it? It would be really easy to do, of course, and the metadata might be useful to others -- you never know. Adding your name, or any other, doesn't lock you in to having to "maintain" it, whatever that means.
- "... determine rules ..." -- why do you need rules? It seems to me it can be as informal as anything else in OSS-land.
- "... maybe you are contributing to a project for the wrong reasons."  Again, all due respect (and I do have a lot of respect for you guys, this is an awesome package) I just think this is a little bit of a backwards way to look at it. I'd prefer to frame it as, "No big deal either way, go ahead and add it." It shouldn't be sticking point to your accepting valuable contributions.

Hope this doesn't upset you -- again, just my two cents. It's your choice, of course!

gagern: This package is part of the npm ecosystem, which decided to employ the `package.json` file as required metadata. So people may expect to find the list of contributors there. The `AUTHORS` file is described as an alternative, and I think that npm will incorporate the information from that into the package description when publishing.

Not having a list of contributors in there at all *may* suggest to some to go looking for that information elsewhere. But having an author and/or some contributors listed makes it look authoritative, thus preventing people from further investigation. So removing all that info from the `package.json` *might* be an option, but it would probably confuse npm and there would still be that copyright holder mentioned in the `LICENSE` file, so I'm not too fond of this approach.

Not everyone is using git and GitHub. So to some people, the files shipped with the released version of the package, as downloaded from npm, are the canonical place to look for such information. One day, one of them may be deciding whether or not to hire me for something…

I think that ideally everyone who contributed significant amounts of work should be mentioned. But I didn't see it as my role to do that in a pull request here, so I only added my own name. In the fork I mantained, I [had added](https://github.com/gagern/libxmljs/commit/46e6d99aa32cd856a7d42f1831cc317f9d2e3f85#diff-b9cfc7f2cdf78a7f4b91a753d10865a2) @defunctzombie, too. I probably should have added some others, too, so I don't claim to properly maintain that list either.

Probably the easiest way to not worry about whom to add is by adding every author git knows about. E.g. `git log --format='%aN <%aE>' | sort | uniq > AUTHORS` and edit to remove different email addresses for the same person. Also remove author and contributors from `package.json` since otherwise those take precedence.

defunctzombie: I think the new maintainers will take care of updating this.

rchipka:
AppVeyor, snprintf macro, etc. all fixed in #407

PR #362, opened 2015-11-16
Fix text merge

teleological:
Extracted from PR #359. Memory fixes when adding merge-able text nodes (implemented in #351) as children or siblings. Includes implementation of XmlText add-sibling methods (already in header). Duplication can be reduced when XmlElement and XmlText (and XmlComment) are refactored to share a base class where methods inappropriate to XmlNode (which is shared by XmlAttribute) can be located.


teleological:
This was broken by #365. Working on a fix.

PR #364, opened 2015-11-18
Ref/Unref closest wrapped ancestor

rc0x03:
Here's a new ref count strategy that works by Ref/Unrefing the closest wrapped ancestor available.

Benefits over #360:
 * We don't have to modify the libxml source code
 * We use only ObjectWrap::Ref/Unref for ref counting

@teleological I would really appreciate a review on this one. I think it may be less bug-prone than #360.

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,3 +38,4 @@ module.exports.SaxPushParser = sax_parser.SaxPushParser;
 
 module.exports.memoryUsage = bindings.xmlMemUsed;
 
+module.exports.nodeCount = bindings.xmlNodeCount;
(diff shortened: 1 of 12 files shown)


teleological:
@rc0x03 Taking a look!

rc0x03:
Hmm, I wonder if there's a way to merge this functionality with #346 so that if we must loop through children to check if there are any wrapped nodes, we only do it when a node is removed. That way we reduce the overhead of checking all the child nodes, attributes, and namespaces for a node that may not even be there. This would keep garbage collection fast for people who use libxmljs primarily for parsing and not document manipulation.

teleological:
Good thought, @rc0x03. I'll take a pass at that.

teleological:
Implemented the merge in #365.

PR #365, opened 2015-12-04
Ref wrapped ancestor plus safe wrapper destruction

teleological:
Merge of #364 and #346: Fixes memory leaks and segfaults. Wrapped nodes maintain reference on nearest wrapped ancestor in order to reduce graph traversal required when destroying wrapper of node in detached graph.

### DIFF
--- a/src/xml_node.cc
+++ b/src/xml_node.cc
@@ -230,66 +230,79 @@ XmlNode::New(xmlNode* node)
 }
 
 XmlNode::XmlNode(xmlNode* node) : xml_obj(node) {
-    this->freed = false;
     xml_obj->_private = this;
+    this->ancestor = NULL;
 
-    // this will prevent the document from being cleaned up
-    // we keep the document if any of the nodes attached to it are still alive
-    XmlDocument* doc = static_cast<XmlDocument*>(xml_obj->doc->_private);
-    doc->ref();
+    // Ref the doc
+    (static_cast<XmlDocument*>(xml_obj->doc->_private))->Ref();
 
-    // check that there's a parent node with a _private pointer
-    // also check that the parent node isn't the doc since we already Ref() the doc once
-    if (xml_obj->parent != NULL &&
-        xml_obj->parent->_private != NULL &&
-        (void*)xml_obj->doc != (void*)xml_obj->parent)
-    {
-        static_cast<XmlNode*>(xml_obj->parent->_private)->Ref();
-    }
+    // ref the nearest wrapped ancestor
+    this->ref_wrapped_ancestor();
 }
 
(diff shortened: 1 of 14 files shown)


teleological:
Please review, @rc0x03 – thanks!

src/xml_node.cc
```diff
@@ -230,66 +230,257 @@ XmlNode::New(xmlNode* node)
...
-        if (parent->refs_ > 0)
-        {
-            parent->Unref();
+    if (xml_obj->parent == NULL) {
+        if (get_wrapped_descendant(xml_obj) == NULL) {
+            xmlFreeNode(xml_obj);
         }
     }
+    else {
+        xmlNode *ancestor = get_wrapped_ancestor_or_root(xml_obj);
```
rc0x03: It might be possible to avoid this lookup if we already know that `this->ancestor` was not null. We could make `this->unref_wrapped_ancestor()` return true if it unrefed an ancestor.

So we could then say something like `hasWrappedAncestor = this->unref_wrapped_ancestor()`, then make the condition `} else if (!hasWrappedAncestor) { ...`. This way we avoid another lookup if we already know about a wrapped ancestor.

rc0x03:
:+1: I think this is the ideal solution. This will keep things fast for parsing/searching and only a little overhead for manipulating documents while still having great segfault/memory leak protection. @teleological thanks for putting so much energy into this (especially the comprehensive tests). I think it will make libxmljs a more viable option for projects that require document creation/manipulation.

src/xml_node.cc
teleological: So I tried that, and the `z_memory_leak` test reported leaked nodes. Reflecting, it makes sense: When we `unref_wrapped_ancestor()`, we know that the unwrapped node was at one time the ancestor of the ref'ing node, but we can't be sure that it's still an ancestor. If the node whose wrapper is being destroyed belongs to a detached graph with no other wrapped nodes, the search that would have discovered this is short-circuited (`ancestor->_private != NULL`) and the graph leaks.

teleological:
@rc0x03 Thank you for your collaboration on this issue, and for starting the ball rolling back in July! @defunctzombie Unless you object, I'd like to merge this along with #362 and #363 with an eye to the next release.

src/xml_node.cc
rc0x03: Ahh that's right, we can't know if the node was removed from its previous ancestor unless we check again. Well that's all I saw as far as possible changes to this code. The rest looks pretty solid.

defunctzombie:
If you both agree this is good, then I have no objections.

rc0x03:
:+1: @teleological looks like you're good to merge once you squash the commits.

PR #366, opened 2015-12-06
0.16.0

teleological:
Changes since 0.15.0:
* Document#type(), Document.fromHtml()
* Node#namespace(ns) returns node
* Node#namespaces(true) returns only locally-defined namespaces
* Comment nodes no longer escape content as XML
* Exposed xmlMemUsed as .memoryUsage()
* Fixed various memory issues

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "Jeff Smick"
   ],
   "description": "libxml bindings for v8 javascript engine",
-  "version": "0.15.0",
+  "version": "0.16.0",
   "scripts": {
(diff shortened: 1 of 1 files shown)


teleological:
@defunctzombie , @rc0x03 Are we ready for release?

@hildjj would you be able to update [the wiki](https://github.com/polotek/libxmljs/wiki/Node) for the new optional argument to `namespaces()`?

defunctzombie:
LGTM, merge, and tag. Then ping me and I will publish.

defunctzombie:
published

PR #373, opened 2015-12-13
Remove previous children when setting element text

teleological:
* Previous children are freed when text is set

### DIFF
--- a/src/xml_element.cc
+++ b/src/xml_element.cc
@@ -420,8 +420,22 @@ XmlElement::get_path() {
 }
 
 void
+XmlElement::unlink_children() {
+  xmlNode *cur = xml_obj->children;
+  while (cur != NULL) {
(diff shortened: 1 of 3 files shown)


rc0x03:
:+1: looks good as long as it works for child attributes and namespaces as well.

teleological:
@rc0x03 Since attributes and namespaces don't descent from XmlElement, this won't affect how they are handled. But as far as I can tell, no special handling is needed. Do you have a case in mind?

rc0x03:
Any child attributes, namespaces, text nodes and comment nodes should also cause a segfault when their parent node text is modified and they try to unref their parent. Look at the attribute example in the test code I provided.

rc0x03:
Actually, since we're unlinking the child it shouldn't matter because we ref the child before we can ref its attributes or namespaces.

Issue #381, opened 2016-01-21
libxmljs@0.17.0 not compiled on windows


suwatch:
libxmljs@0.16.1 compiled fine on Windows.  However, [libxmljs@0.17.0](https://github.com/polotek/libxmljs/commit/9acfcaf7d736ec64656bce04259e3b526ecc4b5d) failed to compile on Windows (for instance, with VS 2015).  One of the errors below.

```
..\..\..\vendor\libxml\catalog.c(26): fatal error C1083: Cannot open include file: 'unistd.h': No such file or directory [D:\home\site\wwwroot\node_modules\libxmljs\build\vendor\libxml\libxml.vcxproj]
```

Grigory90:
Same problem. I back to version 0.16.1

defunctzombie:
@suwatch or @Grigory90 I do not have a windows machine, however if you could create a PR with the include commented out and see if that is the only fix needed I would appreciate that.

suwatch:
@defunctzombie  @ashkandaie thanks for fixing this.  Delivery wise, will this be part of what libxmljs version?

defunctzombie:
I just released it in 0.17.1

Grigory90:
Good job.

suwatch:
0.17.1 built on Windows successfully - lots of warnings though.   They were there even on the earlier versions of libxmljs, fyi.

```
..\..\..\vendor\libxml\catalog.c(979): warning C4013: 'open' undefined; assuming extern returning int [D:\home\site\wwwroot\node_modules\libxmljs\build\vendor\libxml\libxml.vcxproj]        
..\..\..\vendor\libxml\catalog.c(998): warning C4013: 'close' undefined; assuming extern returning int [D:\home\site\wwwroot\node_modules\libxmljs\build\vendor\libxml\libxml.vcxproj]        
..\..\..\vendor\libxml\catalog.c(1005): warning C4013: 'read' undefined; assuming extern returning int [D:\home\site\wwwroot\node_modules\libxmljs\build\vendor\libxml\libxml.vcxproj]        
..\..\..\vendor\libxml\dict.c(188): warning C4244: 'function': conversion from 'time_t' to 'unsigned int', possible loss of data [D:\home\site\wwwroot\node_modules\libxmljs\build\vendor\libxml\libxml.vcxproj]        
[...]

PR #384, opened 2016-02-02
Added preprocessor conditions to config.h to build on windows platforms.

ashkandaie:
In regard to #381. Here is a PR that fixes the build on windows platforms.

### DIFF
--- a/vendor/libxml/config.h
+++ b/vendor/libxml/config.h
@@ -157,7 +157,9 @@
 #define HAVE_RAND 1
 
 /* Define to 1 if you have the `rand_r' function. */
+#ifndef WIN32
 #define HAVE_RAND_R 1
+#endif
 
 /* Define to 1 if you have the <resolv.h> header file. */
 #define HAVE_RESOLV_H 1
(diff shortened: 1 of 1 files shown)


ashkandaie:
Forgot to mention that the parse test in xml_parser fails the last assertion of assert.equal(str, doc.toString());

defunctzombie:
Great!

I am wondering tho how this ever compiled on windows. Looking at the file blame information, it seems that the code in question was the same before and after the recent update of libxml2.

ashkandaie:
Interesting. I'll take a look when time permits.

gagern:
> I am wondering tho how this ever compiled on windows. Looking at the file blame information, it seems that the code in question was the same before and after the recent update of libxml2.

There were some `#undef` lines in `vendor/libxml/libxml.h` which [the update removed](https://github.com/polotek/libxmljs/commit/9acfcaf7d736ec64656bce04259e3b526ecc4b5d#diff-249d2ce676689422cc0e3f3b3c87deac). Something I adressed in https://github.com/gagern/libxmljs/commit/5c328891854ed23ae3c8abcb0f0c3ccfb86d7d3e for #297 which you rejected.

PR #386, opened 2016-02-09
Avoid static library, include libxml objects directly

gagern:
This will result in more of the libxml symbols getting exported from the DLL, so that our node extension becomes a superset of what libxml itself would export. This fixes unresolved external symbol linker errors on windows, described in https://github.com/albanm/node-libxslt/issues/27#issuecomment-156890743.

I already proposed this commit as part of #361 but you wanted multiple pull requests there, so here is one of them. If you can think of a better way to re-export all symbols, without maintaining a list of exported symbols that has to be updated on every version bump of libxml itself, feel free to make suggestions.


gagern:
Travis reports an error for the [`validate_rng_memory_usage` test](https://github.com/gagern/libxmljs/blob/08c56f6d07442622eb72bf52e44e3cd86b43f92a/test/document.js#L365-L406) which appears unrelated. Rerunning the suite should clear this, but perhaps this test warrants some closer inspection in its own right.

defunctzombie:
LGTM

@teleological @rc0x03 thoughts?

rchipka:
Fixed by #454

PR #382, opened 2016-01-22
[WIP] add appveyor.yml config for windows build testing

defunctzombie:

PR #387, opened 2016-02-09
Use AppVeyor configuration from libxmljs-mt

gagern:
You may want to use this instead of #382, since the file here is tested in practice in my fork. Originally a previous version of this file was part of #361 which you wanted split up.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,22 @@
+environment:
+ matrix:
+  - nodejs_version: LTS
+  - nodejs_version: Stable
+
+os:
+  - Visual Studio 2013
+  - Visual Studio 2015
+
+install:
+  - ps: Install-Product node $env:nodejs_version
+  - node --version
+  - npm --version
+  - git submodule update --init --recursive
+
+build_script:
+  - npm install
+
+test_script:
+  - npm test
+
+deploy: off


defunctzombie:
I am all for adding AppVeyor, I've just been waiting on @polotek to setup the hooks cause I am not an admin on this repo.

gagern:
This may be superseded by #407. Comparing with that, this one here

* does not do deployment
* does not test against node prior to version 4
* does test against latest Node versions Stable and LTS, which at some point may be beyond 4 and 5
* does test against Visual Studio 2013 which was important due to different `snprintf` handling (see gagern@7e82427a851d0aae12b0ca39c70f088e717798b7 and gagern@744a48d78ab9a805fbaf4cfd91ef33a7d998a84d) but probably not as important if you do binary releases
* initializes git submodules, which aren't used in the master branch

rchipka:
@gagern Are your AppVeyor tests all passing on libxmljs 0.17.1? When AppVeyor runs the tests in my PR there is some unexpected behavior, so if yours is passing then maybe we should merge this config for 0.17.2 until the test issues are sorted out.

rchipka:
So, @gagern's config [builds successfully](https://ci.appveyor.com/project/rchipka/libxmljs-5pd5d) even on 0.17.1. I'm really not sure what's causing the weird results on my build tests.

rchipka:
Fixed by #407

PR #388, opened 2016-02-10
Baseurl for xml (as well as html)

Klortho:
This PR comprises two changes. I'll describe them as line comments below.

The main purpose of this PR is to add the baseUrl option to FromXml. I noticed that this (and a few other options) were already added to FromHtml a long time ago, so I added them to FromXml in the same way.

The two methods are almost identical, and probably should be refactored to move the common code into a separate function that they both call. If it had been done that way originally, then users of FromXml would have benefited from these extra options added to FromHtml.

So, some of the changes were just for the purposes of harmonizing the FromHtml and FromXml methods. I didn't want to factor out into a separate function, though -- it seemed like too big a change. In lieu of that, I just tried to get the two functions to be as similar as possible.

### DIFF
--- a/src/xml_document.cc
+++ b/src/xml_document.cc
@@ -235,7 +235,7 @@ NAN_METHOD(XmlDocument::ToString)
     if (xmlBufferLength(buf) > 0)
         ret = Nan::New<v8::String>((char*)xmlBufferContent(buf), xmlBufferLength(buf)).ToLocalChecked();
     xmlBufferFree(buf);
-    
+
     return info.GetReturnValue().Set(ret);
 }
 
@@ -317,7 +317,7 @@ xmlParserOption getParserOptions(v8::Local<v8::Object> props) {
     ret |= getParserOption(props, "cdata", XML_PARSE_NOCDATA, false);       // 16384: merge CDATA as text nodes
 
     ret |= getParserOption(props, "noxincnode", XML_PARSE_NOXINCNODE);      // 32768: do not generate XINCLUDE START/END nodes
-    ret |= getParserOption(props, "xinclude", XML_PARSE_NOXINCNODE, false); // 32768: do not generate XINCLUDE START/END nodes
+    ret |= getParserOption(props, "xincnode", XML_PARSE_NOXINCNODE, false); // 32768: do not generate XINCLUDE START/END nodes
 
     ret |= getParserOption(props, "compact", XML_PARSE_COMPACT);            // 65536: compact small text nodes; no modification of the tree allowed afterwards (will possibly crash if you try to modify the tree)
     /*ret |= getParserOption(props, "compact", HTML_PARSE_COMPACT , false);   // 65536: compact small text nodes*/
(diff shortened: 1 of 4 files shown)


src/xml_document.cc
```diff
@@ -317,7 +317,7 @@ xmlParserOption getParserOptions(v8::Local<v8::Object> props) {
     ret |= getParserOption(props, "cdata", XML_PARSE_NOCDATA, false);       // 16384: merge CDATA as text nodes
 
     ret |= getParserOption(props, "noxincnode", XML_PARSE_NOXINCNODE);      // 32768: do not generate XINCLUDE START/END nodes
-    ret |= getParserOption(props, "xinclude", XML_PARSE_NOXINCNODE, false); // 32768: do not generate XINCLUDE START/END nodes
+    ret |= getParserOption(props, "xincnode", XML_PARSE_NOXINCNODE, false); // 32768: do not generate XINCLUDE START/END nodes
```
Klortho: As the commit message describes, I am having trouble getting xinclude to work, and tried to fix it with this change. This *looks* like a typo to me: the `xinclude` property is already defined above, and I think this one should be `xincnode`. But, it didn't fix my problem, and so I'm not sure. If I'm wrong, let me know, and I'll take this out of the PR.

src/xml_document.cc
```diff
@@ -413,19 +416,41 @@ NAN_METHOD(XmlDocument::FromXml)
...
             XmlSyntaxError::PushToArray);
 
-    xmlParserOption opts = getParserOptions(info[1]->ToObject());
+    v8::Local<v8::Object> options = info[1]->ToObject();
+    v8::Local<v8::Value>  baseUrlOpt  = options->Get(
+        Nan::New<v8::String>("baseUrl").ToLocalChecked());
+    v8::Local<v8::Value>  encodingOpt = options->Get(
+        Nan::New<v8::String>("encoding").ToLocalChecked());
+    v8::Local<v8::Value> excludeImpliedElementsOpt = options->Get(
+        Nan::New<v8::String>("excludeImpliedElements").ToLocalChecked());
```
Klortho: `excludeImpliedElementsOpt` isn't used by FromXml, but it is in FromHtml. So, again, the reason for having it here is so that the two functions are the same.

Klortho:
Hello?

Klortho:
In case someone comes along needing this feature, I published this as [@klortho/libxmljs](https://www.npmjs.com/package/@klortho/libxmljs)

defunctzombie:
@teleological @rc0x03 could you take a look?

This also needs tests.

src/xml_document.cc
rc0x03: Yes this is a typo. The `xinclude` option is defined [here](https://github.com/polotek/libxmljs/blob/master/src/xml_document.cc#L305). Good catch.

rc0x03:
Since the `FromXml` and `FromHtml` are so similar I think they should be consolidated into a simpler function. Maybe we could do this in the same PR? If you all agree, maybe @Klortho can make both functions call something like `Parse(void* data, int isHtml = 0)` instead. If not, after tests are added I think this is fine as-is.

Klortho:
I just added a test for baseUrl on fromXml, but

> consolidated into a simpler function. Maybe we could do this in the same PR? 

I'm afraid I don't really have time right now, nor am I confident to do it -- that NAN code looks a little scary.

Klortho:
Tests pass CI.

rchipka:
Tests look good. Squash the commits and we can merge this.

Klortho:
Done

PR #407, opened 2016-05-11
Precompiled binary support

rchipka:
This pull request addresses issue #395 by uploading Windows binaries to GitHub Releases. If a user installs libxmljs on Windows, node-pre-gyp will use one of the precompiled binaries.

This also fixes issue #392 and adds AppVeyor CI testing for Windows builds.

PR #412, opened 2016-05-23
Fix docs for move from polotek/libxmljs to libxmljs/libxmljs

hildjj:
Fix the broken Travis image in the readme, and update the package.json to not require redirects.


rchipka:
Thanks! Fixed by #407

PR #417, opened 2016-06-02
Bump libxml to 2.9.4

gagern:
See [release notes](http://xmlsoft.org/news.html).

This would be so [much easier](https://github.com/gagern/libxmljs/commit/c6fd1a10f1fe4eec2abd3540959325564526a323) with [vendor submodules](https://github.com/libxmljs/libxmljs/pull/297)...


defunctzombie:
@teleological @rchipka

defunctzombie:
While I agree that submodules do prevent having the code in the repo and makes certain updates easier, I have also found that having the code in tree, while making the diffs more noisy, does make it easier for the end user when cloning and building the repo.

rchipka:
@defunctzombie I think I'm in favor of submodules. Although I have had a bad Git submodule config cause issues with NPM installs and Travis CI, I think that can be avoided since @gagern has been doing it successfully for a while.

defunctzombie:
Ok. It is your call!

rchipka:
@defunctzombie I think we should make it work since it seems to be the most popular opinion. Personally, I see the value in having our own copy of the files since it's one less point of an install/build failure, but if it's working smoothly for @gagern without caveats, I really can't see why not.

gagern:
https://github.com/npm/npm/pull/11094 made working with submodules in npm git urls possible. Users of versions released to the npm registry get the full source anyway. Notice how [I'm dual-using my `.gitignore`](https://github.com/gagern/libxmljs/blob/20f75b9d34eb439be45f67c55d216c723de02585/.gitignore#L8-L9) to strip some irrelevant parts out of those, though.

rchipka:
@gagern Thanks! Eventually we'll get around to using submodules.

PR #429, opened 2016-08-17
Fix travis-ci links in README

igorwwwwwwwwwwwwwwwwwwww:
It seems like the repository was moved from the @polotek user to the @libxmljs organization.

This patch changes the link and image to point to the new location.


rchipka:
Was fixed by #407

PR #459, opened 2017-03-02
Load native node module from lib/binding

bwknight877:
This is a fix for Issue [458](https://github.com/libxmljs/libxmljs/issues/458)

These changes move the native node module into `./lib/binding/{node_abi}-{platform}-{arch}/` instead of `./build/Release/`

This is similar to what sqlite3 does, and this is the recommended way according to the node-pre-gyp docs

In binding.gyp, there's an `action_after_build` task that copies the native node module to the correct path specified in package.json

Changes in `lib/bindings.js` now use node-pre-gyp to resolve the path to `xmljs.node` from package.json

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -74,6 +74,17 @@
           },
         }]
       ]
+    },
+    {
+      "target_name": "action_after_build",
+      "type": "none",
+      "dependencies": [ "<(module_name)" ],
+      "copies": [
+          {
+            "files": [ "<(PRODUCT_DIR)/<(module_name).node" ],
+            "destination": "<(module_path)"
+          }
+      ]
     }
   ]
 }
--- a/lib/bindings.js
+++ b/lib/bindings.js
@@ -1,1 +1,5 @@
-module.exports = require('bindings')('xmljs');
+var binary = require('node-pre-gyp');
+var path = require('path');
+var binding_path = binary.find(path.resolve(path.join(__dirname, '../package.json')));
+var binding = require(binding_path);
+module.exports = binding;
--- a/package.json
+++ b/package.json
@@ -6,13 +6,13 @@
   ],
   "binary": {
         "module_name" : "xmljs",
-        "module_path" : "./build/Release/",
+        "module_path" : "./lib/binding/{node_abi}-{platform}-{arch}",
         "host"        : "https://github.com",
         "remote_path" : "./libxmljs/libxmljs/releases/download/v{version}/",
         "package_name": "{node_abi}-{platform}-{arch}.tar.gz"
   },
   "description": "libxml bindings for v8 javascript engine",
-  "version": "0.18.4",
+  "version": "0.18.5",
   "scripts": {
     "install": "node-pre-gyp install --fallback-to-build --loglevel http",
     "test": "node --expose_gc ./node_modules/.bin/nodeunit test"
@@ -31,7 +31,6 @@
   },
   "dependencies": {
     "node-pre-gyp": "~0.6.32",
-    "bindings": "1.2.1",
     "nan": "~2.5.0"
   },
   "devDependencies": {
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 build/
 node_modules/
 npm-debug.log
+lib/binding/


bwknight877:
build and test locally works on my mac

package.json
```diff
@@ -6,13 +6,13 @@
...
         "module_name" : "xmljs",
-        "module_path" : "./build/Release/",
+        "module_path" : "./lib/binding/{node_abi}-{platform}-{arch}",
         "host"        : "https://github.com",
         "remote_path" : "./libxmljs/libxmljs/releases/download/v{version}/",
         "package_name": "{node_abi}-{platform}-{arch}.tar.gz"
   },
   "description": "libxml bindings for v8 javascript engine",
-  "version": "0.18.4",
+  "version": "0.18.5",
```
bwknight877: I'm not sure about your policy to bump the version number, but I had to increment it so that a fresh npm install would not try again to pull it down from github

rchipka:
Will this affect AppVeyor's deployment of prebuilt Windows binaries?

bwknight877:
I'm trying to make sure the output of `node-pre-gyp package` is what we expect with this change

I'm not familiar with AppVeyor's deployment process though

bwknight877:
It looks like the node-pre-gyp package file has the correct contents, but my changes only place the `xmljs.node` file into the package, instead of including the rest of the things from the `build/Release` dir like it did before

image

lib/bindings.js
```diff
@@ -1 +1,5 @@
-module.exports = require('bindings')('xmljs');
+var binary = require('node-pre-gyp');
+var path = require('path');
+var binding_path = binary.find(path.resolve(path.join(__dirname, '../package.json')));
```
rchipka: Is there a cleaner way to do this?

Looks like the bindings module might [automatically detect platform, arch, etc.](https://github.com/TooTallNate/node-bindings/blob/master/bindings.js) depending on the directory structure.

I think if we make the directory structure  `module_root/compiled/version/platform/arch/bindings` then we can just use `require('bindings')()`.

rchipka: Note: Looks like the `compiled` part of the path is either `process.env.NODE_BINDINGS_COMPILED_DIR || "compiled"`

To simplify things we could remove that part by calling `require('bindings')({ compiled: '' })`, which would make our path `module_root/version/platform/arch/bindings`

rchipka:
@bwknight877 Are those missing build artifacts build generated at all?

lib/bindings.js
bwknight877: the original change was straight from node-pre-gyp's documentation
https://github.com/mapbox/node-pre-gyp#3-dynamically-require-your-node

Its the same thing that sqlite3 does:
https://github.com/mapbox/node-sqlite3/blob/master/package.json#L12
https://github.com/mapbox/node-sqlite3/blob/master/lib/sqlite3.js#L1-L5

bwknight877: Telling bindings to look in `module_root/compiled/version/platform/arch/bindings` 
would look for `v6.9.5/win32/x64/xmljs.node`

bindings does some string replacement on those parts of the try path

bwknight877: to keep using bindings, this would also work

```module.exports = require('bindings')({
  bindings: 'xmljs',
  try: [['module_root', 'lib', 'binding', 'node-v' + process.versions.modules + '-' + process.platform + '-' + process.arch, 'bindings']]
});```

Comment (bwknight877) [inline lib/bindings.js]: using node-pre-gyp and `binary.find()` still to me looks like the cleanest way to load xmljs.node from the path specified in package.json

**Comment by bwknight877** [comment]
those missing build artifacts are still generated, but they aren't copied into the directory that is used when you call `node-pre-gyp package`

other than possibly the pdb, I'm not sure you really need to include the rest of the build artifacts, just the xmljs.node and xmljs.pdb

it would be easy enough to add xmljs.pdb to copy during the action_after_build step

File: binding.gyp [hunk 3]
```diff
@@ -74,6 +74,17 @@
...
         }]
       ]
+    },
+    {
+      "target_name": "action_after_build",
+      "type": "none",
+      "dependencies": [ "<(module_name)" ],
+      "copies": [
+          {
+            "files": [ "<(PRODUCT_DIR)/<(module_name).node" ],
```
bwknight877: if you want to include xmljs.pdb, we could add that to copy here

I don't think the other build artifacts are necessary, node certainly doesn't use anything but xmljs.node

lib/bindings.js
rchipka: In my opinion it seems better to let the bindings module handle the finding/loading of the bindings file.

However you want to structure the path is fine as long as all of the build artifacts exist and the binary deployments still work.

binding.gyp
rchipka: I believe those other files are for debugging purposes.

They probably aren't necessary, but if you can find a way to include them that would be good.

lib/bindings.js
rchipka: @bwknight877 I think you're right. Seeing as we'd have to define the binary path in two places using the bindings module, I'd rather just have it in package.json.

It would be nice if the bindings module or node-pre-gyp would automatically do this.

bwknight877: using binary.find() from node-pre-gyp does exactly that, it loads the package.json, and resolves the name and path to xmljs.node from the binary.module_name and binary.module_path defined in package.json

no need to even specify the name `xmljs` or `xmljs.node` because its already defined in package.json

I'm working on a PR for the bindings module to add the node-pre-gyp path like this, no idea if that'll go through

rchipka:
Looks good to me. As long as `node-pre-gyp package` is working properly then AppVeyor should be okay.

lib/bindings.js
rchipka: Sounds good to me.

Alternatively node-pre-gyp's [`binary.find()`](https://github.com/mapbox/node-pre-gyp/blob/master/lib/pre-binding.js#L15) could be modified to accept a JSON object.

Then the code could just be:

```javascript
var bindings = require('node-pre-gyp').find(require('../package.json'));

bwknight877: not sure if it'll get accepted, but I opened a PR against node-bindings to search this path that most native node modules use https://github.com/TooTallNate/node-bindings/pull/32

sdball:
Is there anything blocking this PR from being merged? It's still relevant for apps that do builds for multiple operating systems.

PR #462, opened 2017-03-31
fix to array detection

gerbenmeyer:
An array is an instance of an object, so the check `res instanceof Array` will return false, even in case of an array. Hence, I replaced it with `Array.isArray(res)`. This solved quite a few issues for me :smile:

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -62,7 +62,7 @@ Element.prototype.cdata = function(content) {
 
 Element.prototype.get = function() {
     var res = this.find.apply(this, arguments);
-    if (res instanceof Array) {
+    if (Array.isArray(res)) {
         return res[0];
     } else {
         return res;
@@ -79,4 +79,3 @@ Element.prototype.defineNamespace = function(prefix, href) {
 };
 
 module.exports = Element;
-


defunctzombie:
Thanks for the fix. Could you kindly add a unit test that captures this fix in functionality.

gerbenmeyer:
I didn't manage to make a unit test for this in the setup of this project, however, the issue of `get` returning an Array always happens to me when I'm making unit tests in Jest.

The could be because of what described at https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof, under `instanceof and multiple context (e.g. frames or windows)`.

In either case, using `Array.isArray` is the only way to securely check whether something is an array or not. And this solved all my Jest unit test issues, and all your unit tests are also still passing.

jtinbergen:
Could this fix please be merged, we also keep running into this problem and the fix is really easy (all tests stay green)

nukep:
I'm running into this problem as well when writing tests with Jest. I'm pretty sure this pull request would fix it.
libxmljs version: 0.19.5
Jest version: 23.6.0

To prevent accidental breakage in Jest (and other environments), maybe we can add a comment to warn contributors why we use Array.isArray. We _might_ be able to unit-test this by inspecting the prototype chain or moving the node object across contexts and checking (somehow?), but I'm unsure how to do that - and if it's a lot of work, if it's even worth it (whereas a comment can warn people).

Demonstration:
```javascript
const libxmljs = require('libxmljs')
const doc = libxmljs.parseXml(`<root><child>Hello</child></root>`)

console.log(doc.get('child'))
// Outputs in Node.js repl: {}
// Outputs in Jest: [ {} ]

console.log(doc.get('child').text())
// Outputs in Node.js repl: Hello
// Throws error in Jest: TypeError: doc.get(...).text is not a function



/** testing find(), which get() uses **/
console.log(doc.find('child') instanceof Array)
// Outputs in Node.js repl: true
// Outputs in Jest: false

console.log(Array.isArray(doc.find('child')))
// Outputs in Node.js repl: true
// Outputs in Jest: true
```

nukep:
I had an additional thought. Some pre-ES5 Node.js versions don't support `Array.isArray`. If we're supporting these versions of Node.js, we should make sure it doesn't break there.

marudor:
I need this so I forked the project.
https://github.com/marudor/libxmljs2
(And dropped support for anything below node 8 as 6 will be EOL soon.

hubgit:
Without this fix, `doc.get(…)` in Jest returns an array instead of a single item!

hubgit:
Here's the relevant Jest issue: https://github.com/facebook/jest/issues/2549 and a possible workaround: https://github.com/facebook/jest/issues/2549#issuecomment-423202304

tamtamchik:
I got the same issue here. Gee, why it has not been merged for years? 
It makes `libxmljs` totally unusable when testing with Jest.

PR #475, opened 2017-06-28
Add node 7 & 8 builds

pietermees:
Per https://github.com/libxmljs/libxmljs/issues/474
Not sure if this is all the changes that are needed?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,8 @@ node_js:
 - '0.12'
 - 4
 - 6
+- 7
+- 8
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ os:
 - osx
 env:
 - CXX=g++-4.8
-compiler:		
+compiler:
 - gcc
 - clang
 node_js:
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,6 @@ language: node_js
 os:
 - linux
 - osx
-env:
-- CXX=g++-4.8
-compiler:
-- gcc
-- clang
 node_js:
 - '0.10'
 - '0.12'
@@ -20,13 +15,10 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.8
-    - g++-4.8
-    - g++-4.8-multilib
-    - gcc-multilib
 sudo: false
 before_install:
 - if [[ $TRAVIS_OS_NAME == "linux" ]]; then export CXX=g++-4.8; fi
-- if [[ $TRAVIS_OS_NAME == "osx" ]]; then brew install gcc48; fi
+- if [[ $TRAVIS_OS_NAME == "osx" ]]; then export CXX=clang; fi
 - $CXX --version
 
 - PUBLISH_BINARY=false
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: node_js
 os:
 - linux
 - osx
+env:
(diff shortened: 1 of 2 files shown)


tuananh:
Why not fork it, add to travis and test it yourself?

```
g++-4.8: error: unrecognized command line option '-stdlib=libc++'
```

pietermees:
Eh, yes, I believe that's what this PR is for. As you can see I forked it, made some changes, encountered some build errors that need further fixing, and will try to resolve them.

I noticed that the build failed because of the error you posted. Just haven't found the solution yet. Feel free to let me know if you have any thoughts on how to fix it!

rchipka:
@pietermees @tuananh You can push your changes to this PR until the error is fixed. Just be sure to squash your commits since there are so many.

pietermees:
Thanks, will do!

rchipka:
Seems like the problem is due to the fact that Xcode uses Clang and not GCC, and [gyp is appending](https://chromium.googlesource.com/external/gyp/+/master/pylib/gyp/xcode_emulation.py#676) `-stdlib=libc++` to the `gcc` command even though it's a Clang option.

Two potential solutions:

  1. Use Clang for macOS builds instead of `g++-4.8`. We would also need to be sure to add `-std=c++11` or set `CLANG_CXX_LANGUAGE_STANDARD` to `c++11`.
 2. Use GCC, but somehow tell gyp that we're not using Clang and not to append Clang-specific args.

pietermees:
Agreed!
I did notice that we include
```
compilers:
- gcc
- clang
```
Which seems to indicate we want to use both? Was there a specific reason for that?

If there's no need to support both compilers, it seems easier to just compile with clang instead of gcc. Do you agree?

rchipka:
Yeah if it will build with Clang for all versions of macOS then we should just use Clang.

I think that gyp will probably want to use GCC on linux by default unless we can override it.

Maybe we could use GCC on linux and Clang on macOS?

```
if [[ $TRAVIS_OS_NAME == "linux" ]]; then export CXX=g++-4.8; fi
if [[ $TRAVIS_OS_NAME == "osx" ]]; then export CXX=clang; fi
```

pietermees:
Hmm, seems like the osx backlog on Travis is making it difficult to debug right now. I'll check further in the morning ;)

rchipka:
Good to merge?

pietermees:
I was still cleaning a few small things up, because Travis complains during the build about having a C++11 compiler available (like you mentioned earlier).

I don't think it's a big deal though, as I think it's resolved when we set the proper CXX export later in the script. So their warning is probably irrelevant.

rchipka:
Sorry, if the changes are small then I'll just put them in the 0.18.7 update.

Where did you see the message? I can't find it in the Travis logs.

I'm not sure if we should put the c++11 flag in the travis config or `binding.gyp`.

pietermees:
See https://travis-ci.org/libxmljs/libxmljs/jobs/248162828

Is the flag needed? From https://docs.travis-ci.com/user/languages/javascript-with-nodejs#Node.js-v4-(or-io.js-v3)-compiler-requirements I gathered that simply adding
```
env:
  - CXX=g++-4.8
```
might solve it. It was there before, but I removed it since we were overriding it in the `before_install` script anyway. Putting it back and removing the line `- if [[ $TRAVIS_OS_NAME == "linux" ]]; then export CXX=g++-4.8; fi` might be ok?

pietermees:
Like this: https://github.com/pietermees/libxmljs/blob/patch-1/.travis.yml

PR #478, opened 2017-07-05
Document find in the front README

madarche:
This is important to document the `find()` method right in the front README, it took me some time to find out how to have a collection of elements through XPath. I was getting mad at the `get()` method, before finding the wiki being well documented. Moreover `childNodes()` is documented, so let's do the "equivalent" for XPath.

FYI I have documented a bit more how to use `find()` in https://github.com/libxmljs/libxmljs/wiki/Element#elementfindxpath-namespaces

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,25 +6,31 @@
 LibXML bindings for [node.js](http://nodejs.org/)
 
 ```javascript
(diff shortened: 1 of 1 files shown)


madarche:
Anyone to merge this very simple PR?

PR #482, opened 2017-07-31
Added encoding for toString() method

### LINKED ISSUE #398: xml entities are not preserved
In our team we encode all space characters other than plain space as entities so you can tell one from another. But they aren't preserved by read/serialize. Is it a bug or expected behavior? Is there any way to preserve them?

``` js
var libxmljs = require("libxmljs");

var doc = libxmljs.parseXml("<x>&#160;</x>");
console.log(doc.toString());  // outputs '<x> </x>' while original '<x>&#160;</x>' wanted
```

rhodgkins:
Need to specify the encoding as `HTML` to fix #398 

(See http://xmlsoft.org/encoding.html#Default)

This PR adds the ability to provide a `encoding` option to `toString()`.

### DIFF
--- a/test/html_parser.js
+++ b/test/html_parser.js
@@ -140,5 +140,13 @@ module.exports.toString = function(assert) {
     assert.ok(doc.toString({ type: 'xml' }).indexOf('<?xml') > -1);
     assert.ok(doc.toString({ type: 'xhtml' }).indexOf('<?xml') > -1);
     assert.ok(doc.toString({ type: 'xml', selfCloseEmpty:true }).indexOf('<a/>') > -1);
+
+    doc = libxml.parseHtml("<a>Something&nbsp;with a space</a>");
+    assert.ok(doc.toString().indexOf('<?xml') === -1);
+    assert.ok(doc.toString({ type: 'xml' }).indexOf('<?xml') > -1);
+    assert.ok(doc.toString({ type: 'xhtml' }).indexOf('<?xml') > -1);
+    assert.ok(doc.toString({ type: 'xhtml' }).indexOf('&nbsp;') === -1);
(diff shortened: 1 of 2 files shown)


test/html_parser.js
```diff
@@ -140,5 +140,12 @@ module.exports.toString = function(assert) {
     assert.ok(doc.toString({ type: 'xml' }).indexOf('<?xml') > -1);
     assert.ok(doc.toString({ type: 'xhtml' }).indexOf('<?xml') > -1);
     assert.ok(doc.toString({ type: 'xml', selfCloseEmpty:true }).indexOf('<a/>') > -1);
+
+    doc = libxml.parseHtml("<a>Something&nbsp;with a space</a>");
+    assert.ok(doc.toString().indexOf('<?xml') === -1);
+    assert.ok(doc.toString({ type: 'xml' }).indexOf('<?xml') > -1);
+    assert.ok(doc.toString({ type: 'xhtml' }).indexOf('<?xml') > -1);
+    assert.ok(doc.toString({ type: 'xhtml' }).indexOf('&nbsp;') === -1);
+    assert.ok(doc.toString({ type: 'xhtml', encoding: 'HTML' }).indexOf('&nbsp;') > -1);
```
rchipka: What does `encoding: 'HTML'` actually do? Isn't this supposed to be a character encoding?

rchipka: Maybe also add a test for encoding a UTF-8 character to ASCII?

rhodgkins: Will do!

Any ideas why its failing?

rhodgkins:
I've updated this change to the latest code from master but not sure why its failing on OS X with Node > 7...!

rchipka:
Thanks!

PR #493, opened 2017-10-26
Provide bindings for the Text Writer API

znerol:
Support basic set of [xmlwriter](http://xmlsoft.org/html/libxml-xmlwriter.html) interface of libxml2. This PR has been proposed in #80 then migrated to #91 and was rejected in #104. Since then I've been maintaining this branch in my own repo. This is because [xmlshim](https://www.npmjs.com/package/xmlshim) depends on the xmlwriter API.

Now that libxmljs is providing binaries, maintenance of the fork has proven to be even more difficult than before :/

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -24,6 +24,7 @@
         'src/xml_syntax_error.cc',
         'src/xml_text.cc',
         'src/xml_pi.cc',
+        'src/xml_textwriter.cc',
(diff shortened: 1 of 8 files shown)


nevercast:
This needs to be reconsidered. Maintaining a fork isn't viable now that binaries are being distributed.

z-ax:
Any news on this?

PR #505, opened 2018-04-26
Build for Node 9 and Node 10

pietermees:


pietermees:
I also wonder whether Node 0.10 and 0.12 should be deprecated.

tamtamchik:
Having the same issues.

rchipka:
@pietermees Between this or #507, I'll except a PR that skips the test, but only if it just skips for the problematic NodeJS version and prints a big warning for future reference. 

This will be good enough for now so that people can still get binaries. We'll sort out any memory issues in the future.

rchipka:
Fixed by #508

PR #507, opened 2018-05-09
Patch which allows to pass build for node 10 and 9

kbiedrzycki:
This patch introduces new versions of dependencies and modifies build pipeline a bit, to make this library work with node 10. I had to ignore one of the specs, maybe someone has an idea how to actually make it work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,12 @@ os:
 - linux
 - osx
 node_js:
-- '0.10'
-- '0.12'
 - 4
 - 6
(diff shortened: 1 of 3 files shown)


rchipka:
@kbiedrzycki I will merge this with the test skipped, but you'll need update the code to only skip the test if `process.version` matches the failing NodeJS version, and if it does you should `console.log` a big warning for future reference.

rchipka:
Fixed by #508

PR #508, opened 2018-05-10
enable node 5 and 9 in travis

tuananh:
enable node 5, 9 and 10 in travis

### DIFF
--- a/test/memory_management.js
+++ b/test/memory_management.js
@@ -1,8 +1,12 @@
 var libxml = require('../index');
+var semver = require('semver');
 
 if (!global.gc) {
     throw new Error('must run with --expose_gc for memory management tests');
 }
+
+var nodeVersion = process.versions.node
+var shouldSkip = semver.satisfies(nodeVersion, '8.x || 9.x || 10.x')
 
 module.exports.setUp = function(done) {
     collectGarbage();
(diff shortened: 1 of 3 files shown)


tuananh:
@rchipka there was a couple of failures , always on these 2 tests

- inaccessible_tree_freed
- inaccessible_document_freed_when_node_freed

Any idea why these tests are so flaky?

rchipka:
@tuananh you could try to update the packages to the same versions in #507

test/memory_management.js
```diff
@@ -20,6 +24,10 @@ module.exports.inaccessible_document_freed = function(assert) {
 };
 
 module.exports.inaccessible_document_freed_when_node_freed = function(assert) {
+    if (isNode8Or9) {
+        assert.done()
```
rchipka: We should `console.log` a big warning message if we're going to do this

tuananh: lemme try to update deps and see if it goes away first

rchipka: Looks like it's still broken. I'm not sure how since [node 8 passes](https://travis-ci.org/libxmljs/libxmljs/jobs/376790433) on #507, even for `inaccessible_tree_freed`.

tuananh:
@rchipka I still got one failure

```
/Users/travis/.node-gyp/0.12.18/include/node/v8.h:5800:54: error: 'CreateHandle' is a protected member of 'v8::HandleScope'
```
in node `0.12`

Do you know how to get this fixed? On `0.10` it's ok though.

I suspect the deps but i rolled it back already.

tuananh:
looks like i have to edit v8 code moving `'CreateHandle' is a protected member of 'v8::HandleScope'` to public section so i disabled node 0.10 and 0.12. They are too old btw.

rchipka:
@tuananh Thanks!

PR #521, opened 2018-07-16
Add libxml2 processing instructions support

### LINKED ISSUE #515: Add the ability to insert processing instructions
As part of my build process, I want to add processing instructions to some of my `svg` files, for example:

```xml
<?xml-stylesheet href="/style.css" type="text/css"?>
```

Unfortunately, `libxmljs` does not currently have such an API. From my brief research, `libxml2` does have an API to do this - `xmlTextWriterStartPI` and `xmlTextWriterEndPI`.
Some more info: https://www.linuxquestions.org/questions/programming-9/how-to-add-xml-stylesheet-tag-in-a-xml-file-using-libxml2-946645/

Thanks for the support!

perrin4869:
Added a bridge to `xmlNewDocPI` for the ability of inserting and manipulating xml processing instructions. 
Closes #515

Adds a new `ProcessingInstruction` top level export class, which shares most of its implementation with the `Comment` class. The only difference is that the ProcessingInstruction also has a `name` attribute

Usage examples:

1) Adds a processing instruction before the root node

```js
const libxmljs = require('libxmljs');

var xml =  '<?xml version="1.0" encoding="UTF-8"?>' +
           '<root>' +
               '<child foo="bar">' +
                   '<grandchild baz="fizbuzz">grandchild content</grandchild>' +
               '</child>' +
               '<sibling>with content!</sibling>' +
           '</root>';

var xmlDoc = libxmljs.parseXml(xml);
var pi = libxmljs.ProcessingInstruction(xmlDoc, 'mypi', 'mycontent');
xmlDoc.root().addPrevSibling(pi);
console.log(xmlDoc.toString());
/*
<?xml version="1.0" encoding="UTF-8"?>
<?mypi mycontent?>
<root>
  <child foo="bar">
    <grandchild baz="fizbuzz">grandchild content</grandchild>
  </child>
  <sibling>with content!</sibling>
</root>
*/
```

2) `ProcessingInstruction` is a node, so it can be manipulated like any other node, in this case, changing the contents:

```js
const libxmljs = require('libxmljs');

var xml =  '<?xml version="1.0" encoding="UTF-8"?>' +
           '<?mypi mycontent?>' +
           '<root>' +
               '<child foo="bar">' +
                   '<grandchild baz="fizbuzz">grandchild content</grandchild>' +
               '</child>' +
               '<sibling>with content!</sibling>' +
           '</root>';

var xmlDoc = libxmljs.parseXml(xml);
console.log(xmlDoc.root().prevSibling().text()); // mycontent
xmlDoc.root().prevSibling().text('mynewcontent');
console.log(xmlDoc.root().prevSibling().text()); // mynewcontent
```

Need to update the wiki

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -23,6 +23,7 @@
         'src/xml_sax_parser.cc',
         'src/xml_syntax_error.cc',
         'src/xml_text.cc',
+        'src/xml_pi.cc',
         'src/xml_xpath_context.cc',
         'vendor/libxml/buf.c',
         'vendor/libxml/catalog.c',
--- a/index.js
+++ b/index.js
@@ -24,6 +24,7 @@ module.exports.features = bindings.features;
 module.exports.Comment = require('./lib/comment');
 module.exports.Document = Document;
 module.exports.Element = require('./lib/element');
+module.exports.ProcessingInstruction = require('./lib/pi');
 module.exports.Text = require('./lib/text');
 
 // Compatibility synonyms
--- a/lib/element.js
+++ b/lib/element.js
@@ -3,7 +3,7 @@ var bindings = require('./bindings');
 /// create a new element on the given document
 /// @param doc the Document to create the element for
 /// @param name the element name
-/// @param {String} [contenn] element content
+/// @param {String} [content] element content
 /// @constructor
 function Element(doc, name, content) {
     if (!doc) {
--- a/lib/pi.js
+++ b/lib/pi.js
@@ -1,0 +1,23 @@
+var bindings = require('./bindings');
+
+/// create a new processing instruction on the given document
+/// @param doc the Document to create the processing instruction for
+/// @param name the processing instruction name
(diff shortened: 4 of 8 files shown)


perrin4869:
Edit: Added more explanations and code samples

rchipka:
LGTM as long as no memory leaks are added to existing APIs

rchipka:
Not exactly sure what function a ProcessingInstruction has either, but if it's in libxml it must be a thing.

perrin4869:
Should have no effect on existing APIs.
Processing instructions are standard XML afaik. From wikipedia, they are "intended to carry instructions to the application", and the most common use case is to embed stylesheets in svg files, as in `<?xml-stylesheet type="text/xsl" href="style.xsl"?>`

rchipka:
@perrin4869 Ah. Are people still using XML? lol

perrin4869:
I hear they still do indeed :P

PR #568, opened 2020-02-25
Update array check

### LINKED ISSUE #566: Array "instanceof" issue in libxml.Document && libxml.Element get function.
Hey, **LibxmlJS team**,
Due to the issue of `instanceof Array` in nodeJS (Link: https://github.com/facebook/jest/issues/2549), some times `get` function returns an array, instead of `array[0]`. After a couple of researches, I found out, `instanceof Array` in `get` function doesn't work properly in several cases and I'd changed it to `Array.isArray(res)`, now it works fine. 

**Examle of fixed issue code:**
```
libxml.Document.prototype.get = function() {
    const res = this.find.apply(this, arguments);
    if (Array.isArray(res)) {
        return res[0];
    } else {
        return res;
    }
};
```
Could you look at this issue and tackle this problem, due others wouldn't have this problem. 

Best regards, **fayzzz**

ktilcu:
`instanceof` can be misleading at times. Specifically, when dealing with
arrays created in different contexts, the array Prototype will be
different allowing an array to be returned from `.get` calls.

This change uses `Array.isArray` as suggested by #566.

Closes #566

### DIFF
--- a/lib/element.js
+++ b/lib/element.js
@@ -62,7 +62,7 @@ Element.prototype.cdata = function(content) {
 
 Element.prototype.get = function() {
     var res = this.find.apply(this, arguments);
-    if (res instanceof Array) {
+    if (Array.isArray(res)) {
         return res[0];
     } else {
         return res;


marudor:
Duplicate of https://github.com/libxmljs/libxmljs/pull/462
As this repository is mainly dead I forked it a while ago. This is fixed in my fork (as well as several other things)
https://github.com/marudor/libxmljs2

PR #600, opened 2022-06-13
Add ARM64 build job on the upload 

### LINKED ISSUE #598: ARM64 build for libxmljs
This repo doesn't have a build for the ARM64 arch, but we don't want to build this binary on the client side. Is it possible to create a build for this? I know this is not a trivial task since Github doesn't support these kind of runners, but I could give it a go. Any hint?

nunomarks:
Closes #598.

@rchipka please let me know what you think!


nunomarks:
Ping @rchipka

rchipka:
@nunomarks Thanks!

PR #635, opened 2023-10-05
chore: add build steps for node 20

anderssonoscar0:
This PR adds build steps for node 20.

It also contains the followin changes:

1. Splits the macos and windows builds since the MAC build didn't like the `npm config set msvs_version "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise" --global` argument into npm.

2. Removes the "invalid"? npm config command `npm config set spin false` which no longer is supported in newer NPM versions.

3. Adds builds steps arm64 macos

### DIFF
--- a/.github/workflows/test-deploy.yml
+++ b/.github/workflows/test-deploy.yml
@@ -7,7 +7,7 @@ on:
     workflow_dispatch:
 
 jobs:
-    upload:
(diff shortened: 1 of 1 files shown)


anderssonoscar0:
@rchipka Could you take a look at this one when you have time? 🙏

PR #639, opened 2023-10-18
Fix ci darwin arm64

pahann:
Hi @rchipka @anderssonoscar0 

this PR in order to fix https://github.com/libxmljs/libxmljs/issues/637 (basically removed darwin arm64 build)

Today the crossbuild darwin x64_86 -> arm64 seems not to be supported on standard x86_64 macos runners, thus the supposed arm64 artefact (https://github.com/libxmljs/libxmljs/releases/download/v1.0.10/node-v115-darwin-arm64.tar.gz) contains a x86_64 file, and this one is not usable on a M1/M2 cpu.

An option would be to use a recently released github silicon worker (https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/) but they are not available for free by contrast to standard macos runners.

That's why I suggest to remove this build for now.

### DIFF
--- a/.github/workflows/test-deploy.yml
+++ b/.github/workflows/test-deploy.yml
@@ -49,48 +49,6 @@ jobs:
               with:
                   asset_paths: '["./build/stage/libxmljs/libxmljs/releases/download/**/*.tar.gz"]'
 
-    upload-macos-arm64:
-        strategy:
-            fail-fast: false
-            matrix:
-                config:
-                    - { node-version: 8, npm-version: 6 }
-                    - { node-version: 9, npm-version: 6 }
-                    - { node-version: 10, npm-version: 6 }
-                    - { node-version: 11, npm-version: 6 }
-                    - { node-version: 12, npm-version: 8 }
-                    - { node-version: 13, npm-version: 8 }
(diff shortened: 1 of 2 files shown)


anderssonoscar0:
@rchipka Seems like parts of my PR, add buildsteps for node 20 broke it for macos 😞 

Would be nice if there was a fix instead of removing this 🙏

pahann:
@anderssonoscar0 maybe there is. But I've only seen the way with not-free macos silicon runners (I've not tested it), there are maybe also solutions with external runners, but I'm too novice in the world of github actions to give a better answer 😕. 

The thing is, right now, this version is broken if we try to use it on M1/M2 😕

qraynaud:
The node20 build is working and a good thing. Having `lastest` on broken build for some does seem bad. I think this should be merged.

rchipka:
Thanks

pahann:
@rchipka thanks a lot. Some builds didn't work, is it a known issue ? 
https://github.com/libxmljs/libxmljs/actions/runs/6562349024/job/17824106736

It doesnt look related to the removal i've done

PR #641, opened 2023-11-08
Bugfix/640

### LINKED ISSUE #640: nextSibling/prevSibling always returns XMLNode
The following methods always return an XMLNode: 

- `nextSibling`
- `prevSibling`
- `prevElement`
- `nextElement`

The reason for this is that the prevSibling and nextSibling methods call the `createXMLReference` method and not `refToNodeType`:
https://github.com/libxmljs/libxmljs/blob/3121013c0e447d267bf4514317c52ad81670f2f5/lib/node.ts#L433 
https://github.com/libxmljs/libxmljs/blob/3121013c0e447d267bf4514317c52ad81670f2f5/lib/node.ts#L437

kpalatzky:
Fixes #640

### DIFF
--- a/lib/node.ts
+++ b/lib/node.ts
@@ -429,12 +429,12 @@ export class XMLNode extends XMLReference<xmlNodePtr> {
         return children;
     }
 
-    public prevSibling() {
-        return refToNodeType(this.getNativeReference().prev);
+    public prevSibling(): XMLNode|null {
+        return refToNodeType(this.getNativeReference().prev) as XMLNode|null;
     }
 
-    public nextSibling() {
-        return refToNodeType(this.getNativeReference().next);
+    public nextSibling(): XMLNode|null {
+        return refToNodeType(this.getNativeReference().next) as XMLNode|null;
     }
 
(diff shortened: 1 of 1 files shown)


kpalatzky:
@rchipka The pipeline fails due to a GYP error. I cannot fix this error myself.

rchipka:
We shouldn't use refToNodeType because then we have to use a type assertion.

Instead we should check if the `.prev` ref is null and return null, otherwise return `createXMLReference(XMLNode,...)`, that way the return type will be inferred correctly.

rchipka:
Also if you could add a test for this that would be greatly appreciated

rchipka:
Hmm, although that does raise the question of how to handle casting to XMLElement in nextElement/prevElement

kpalatzky:
It also seems that the build on ubuntu does work fine. Whats reason why it does fail on macos? I am not familiar with GYP errors.

rchipka:
Seems like maybe a python version issue considering the error is related to unknown flags for an `open()` call

rchipka:
I agree that we should always return an instance that best matches the underlying node ref type, however I'm wondering if there's a way we can do this without a manual type assertion

Issue #628, opened 2023-06-15
Question about `Document.validate`


sbalaji1996:
I have a decently sized xsd file (~7 mb) that I'm using to validate various xml documents that I'm constructing. My code looks something like this

```
async function validateXML(xmlString) {
    const xsdString = (await fs.readFile(xsdFilePath)).toString();
    const xsd = libxmljs.parseXml(xsdString);
    const xml = libxmljs.parseXml(xmlString);
    xml.validate(xsd);
}
```

I'm finding that after I invoke `validate`, the majority of the time spent during validation (a few hundred milliseconds) is in the `xmlSchemaParse` function here, based on some console statements that I've added to test performance: https://github.com/libxmljs/libxmljs/blob/650733271b7ed60b473a76b19791075d55e6e77b/lib/document.ts#L267

My question is - if the xsd file is static, is there a way to provide the library that file (or a reference to that file) _once_ and just have it reference that instead of attempting to parse the xsd every time I invoke `validate`?

rchipka:
There isn't currently a way, but we could provide a function to parse and return schema documents then overload Document.validate to accept an already-parsed schema document

axel-capodaglio:
Hi, is there any update on this?

CCheminaud:
I have the same need on my side, here is a PR: https://github.com/libxmljs/libxmljs/pull/656.
Probably it could be improved, I'm open to suggestions.

PR #656, opened 2024-04-10
Validate multiple doc without rebuilding schema

CCheminaud:
Proposal for #628.

The goal is to provide a way to validate multiples docs without parsing again the schema.

```typescript
import libxmljs from "libxmljs";

const xsd = '<?xml version="1.0" encoding="UTF-8"?>...'

// Parse schema and build context
const schema = libxmljs.parseSchema(xsd)

// Validate all docs
schema.validateDocument(xmlDoc1)
schema.validateDocument(xmlDoc2)
schema.validateDocument(xmlDoc3)

// Still working
xmlDoc1.validate(xsd)
```

### DIFF
--- a/.github/workflows/test-merge.yml
+++ b/.github/workflows/test-merge.yml
@@ -16,6 +16,10 @@ jobs:
                     - { os: windows-2019, arch: x64 }
         steps:
             - uses: actions/checkout@v3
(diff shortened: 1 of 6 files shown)


rchipka:
Nice work, thanks!!

PR #668, opened 2025-03-13
fix: make parseXMLAsync and parseXML have the same defaults for encoding

cranberryofdoom:
# Context

We're currently using `libxmljs` to parse XML for SAML IDP metadata. When we fetch the metadata from the service, the `response.body` may sometimes come back as a string with the following invisible first character: `\uFEFF`.

This character is the Unicode Byte Order Mark (BOM). It’s a special invisible character that's often used at the beginning of text files to indicate the byte order (endianness) of the file, and it is particularly common for UTF-8 encoded files.

I noticed that the behavior between `libxmljs.parseXml` and `libxmljs.parseXmlAsync` handled this invisible character differently. 

`libxmljs.parseXml` succeeds with no issue because if the `buffer` is a `string`, it will default to an encoding of `UTF-8`.

However, `libxmljs.parseXmlAsync` fails because it only uses `DEFAULT_XML_PARSE_OPTIONS.encoding`. This is not defined, which then it throws with the following error: `Error: Start tag expected, '<' not found`.

# The Change
`libxmljs.parseXml` and `libxmljs.parseXmlAsync` both now have the following as its encoding argument: `options.encoding || DEFAULT_XML_PARSE_OPTIONS.encoding || (typeof buffer === "string" ? "UTF-8" : null),` which should handle this invisible character case.

### DIFF
--- a/lib/parse.ts
+++ b/lib/parse.ts
@@ -229,7 +229,7 @@ export const parseXml = (buffer: string | Buffer, options: XMLParseOptions = DEF
             buffer,
             typeof buffer === "string" ? Buffer.byteLength(buffer) : buffer.length,
             options.baseUrl || DEFAULT_XML_PARSE_OPTIONS.baseUrl || "",
-            options.encoding || (typeof buffer === "string" ? "UTF-8" : null),
+            options.encoding || DEFAULT_XML_PARSE_OPTIONS.encoding || (typeof buffer === "string" ? "UTF-8" : null),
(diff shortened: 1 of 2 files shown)


rchipka:
Nice work, thanks!!

elb-notion:
Hey @rchipka, will there be a new release any time soon? Would love to get some of the recent updates in without having to import the github repo instead of the package

